PR #1, opened 2015-11-01
Prototype app

bassosimone:
This pull request introduces the prototype Android app. There are still number of limitations, such as that currently tests cannot run in parallel. However this is already used by me on the emulator as well as on a real Android device (Nexus 7 using Android 5.1.1).

This pull request currently contains also the *devpack* for Android generated using MeasurementKit. I've not figured out yet whether this is the optimal way to bundle it with the application (I dislike adding binary libraries to master because they weight quite a lot -- for this reason for now I've included only x86 and armeabi, rather than including all the available architectures).

- - -

Note: the `io/github/measurement_kit/{common,ooni}` dirs are copied from measurement-kit's main repository. I don't know if this is the best approach. Open-peer provided the `.class` files, but I thought providing the sources was more transparent to the user.


bassosimone:
Closing in favor of #2.

PR #2, opened 2015-12-09
Prototype Android application (try 2)

bassosimone:
This prototype Android application is based on the new repository hosting Measurement Kit JNI libs available at:

    https://github.com/bassosimone/measurement-kit-jni-libs

The implementation uses the JNI libs and their interface to run in synchronous mode OONI tests. The basic mechanism to run in background (inside a so-called service) and to notify back the main thread when work is done is already implemented.

Compared to #1, this is significantly simpler in terms of JNI code and interfaces involved.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,7 @@
 .DS_Store
 /build
 /captures
+
+app/src/main/java/io/github/measurement_kit/jni/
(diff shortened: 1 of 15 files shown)


bassosimone:
This has been tested on a Nexus 5 with Android 6. Merging.

Issue #4, opened 2016-01-12
Find a reliable way to get the device's dns


bassosimone:

bassosimone:
Closed via #8.

PR #8, opened 2016-02-16
Adding a method to get DNS from device

Lymphatus:
The `getDNS()` method asks the device fro DNS resolvers, both for mobile and WiFi. If it fails for some reason, the app will just use the Google DNS instead.

Tested on API 22 (device) and 23 (emulator).

Also changed the compile SDK and target to API 23 and build-tools to 23.0.1.

### DIFF
--- a/app/src/main/java/io/github/measurement_kit/app/MainActivity.java
+++ b/app/src/main/java/io/github/measurement_kit/app/MainActivity.java
@@ -15,6 +15,9 @@ import android.util.Log;
 
 import java.io.FileOutputStream;
 import java.io.InputStream;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.ArrayList;
 
 import io.github.measurement_kit.jni.DnsApi;
 import io.github.measurement_kit.jni.LoggerApi;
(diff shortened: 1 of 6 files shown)


Lymphatus:
Also, shouldn't the app get DNS every time it runs a test, or some of them? They may change since the start (e.g. switching to mobile/WiFi or moving around).

bassosimone:
@Lymphatus, I agree in using the most recent DNS gathered just before the test runs. Since this is also an issue for iOS, I think the best approach is to fix this at the MK level. I've created an issue for this in MK repository: measurement-kit/measurement-kit#331.

app/src/main/java/io/github/measurement_kit/app/MainActivity.java
```diff
@@ -153,5 +161,31 @@ private void copyResources() {
...
+            Method method = SystemProperties.getMethod("get", String.class);
+
+            for (String name : new String[]{"net.dns1", "net.dns2", "net.dns3", "net.dns4",}) {
+                String value = (String) method.invoke(null, name);
+                if (value != null && !value.equals("") && !servers.contains(value)) {
+                    servers.add(value);
+                }
+            }
+        // Using 4 branches to show which errors may occur
+        // We can just catch Exception
```
bassosimone: I really like this that all the possible errors are enumerated!

bassosimone:
Closes #8.

bassosimone:
Tested, thank you! :+1:

bassosimone:
Actually it closes #4.

PR #9, opened 2016-04-29
App v0.1.0 

lorenzoPrimi:
- GUI totally redesigned and improved
- Background tests with AsyncTask
- Different logs for different tests
- Renamed package
- Re-structured project


bassosimone:
Very good work!

Before merging, I'd like to see the following changes applied:

- [ ] like we did for the `-android` repository, you should edit `README.md` telling the user that packages could also be signed using your GPG key

- [ ] could you add an `XXX` comment before the place where we use `AsyncTask` stating that this may not be the optimal solution since OONI tests could take a long time to complete?

bassosimone:
Thank you, merging!

PR #10, opened 2016-06-08
Porting of app to v0.2.0

bassosimone:
In this branch we'll working on porting the app to v0.2.0 and testing it. At the moment of creating this pull request, the ported app was tested in the emulator with many things to fix before merging is possible.


app/src/main/java/org/openobservatory/netprobe/activity/MainActivity.java
```diff
@@ -201,8 +205,10 @@ private void executeTest(int test){
     @Override
     public void update(Observable observable, Object data) {
         if (mTestsListAdapter != null) {
-            mTestsListAdapter.setData(TestData.getInstance().mNetworkMeasurementsRunning);
-            mTestsListAdapter.addData(TestData.getInstance().mNetworkMeasurementsFinished);
+            //mTestsListAdapter.setData(TestData.getInstance().mNetworkMeasurementsRunning);
+            //mTestsListAdapter.addData(TestData.getInstance().mNetworkMeasurementsFinished);
+            mTestsListAdapter.setData(ts.loadTests(this));
+            //mTestsListAdapter.addData(ts.loadTests(this));
```
bassosimone: I think we should cleanup the commented code before merging this branch

app/src/main/java/org/openobservatory/netprobe/model/NetworkMeasurement.java
```diff
@@ -5,11 +5,18 @@
...
-    public NetworkMeasurement(String name, String file){
+    public final String json_file;
+    public final String log_file;
+    //public final String status;
+
+    public NetworkMeasurement(String name){
         this.testName = name;
-        this.fileName = file;
+        this.test_id = System.currentTimeMillis();
+        this.log_file = "/test-"+ test_id +".txt";
```
bassosimone: Super bikesheddish: I'd name this with `.log` extension

app/src/main/java/org/openobservatory/netprobe/utils/Alert.java
```diff
@@ -52,4 +52,57 @@ public void onClick(DialogInterface dialog, int id) {
 
                 }).show();
     }
+
+    public static String formatString(String text){
```
bassosimone: Uhm, this is to pretty-indent the JSON right? I'd see if it's possible to do this task using `Gson`.

lorenzoPrimi: Yes you are right.

Makefile
```diff
@@ -4,12 +4,12 @@ PHONIES += check download-and-verify help unpack unpack-clean
...
 
-REPO      = measurement-kit/measurement-kit-jni-libs
+REPO      = measurement-kit/measurement-kit
 BASEURL   = https://github.com/$(REPO)/releases/download
 PACKAGE   = org/openobservatory/measurement_kit
 SRCDIR    = app/src/main
-TAG       =
-VERSION   = v0.1.2.2
+TAG       = -0
+VERSION   = v0.2.0-beta.2
```
bassosimone: Uhm, the version tagged in here is not super up to date. I'm recompiling everything and pushing `0.2.7` compiled for Android such that we can use the real latest version.

app/src/main/java/org/openobservatory/netprobe/activity/MainActivity.java
```diff
@@ -42,25 +42,31 @@
...
         mTestsListView.setLayoutManager(new LinearLayoutManager(this));
+        mTestsListAdapter.setData(ts.loadTests(this));
 
         TestData.getInstance().addObserver(this);
 
         Button button;
 
+        // XXX: Code commented out; now this must be performed before each test and
+        // the DNS server should be supplied to the test itself
+        /*
```
bassosimone: @lorenzoPrimi do we now query the device for DNS before each test? If so, I'd remove the commented out code, otherwise I'd push a patch for it.

app/src/main/java/org/openobservatory/netprobe/data/TestData.java
```diff
@@ -20,28 +21,28 @@
 public class TestData extends Observable {
     private static final String TAG = "TestData";
 
-    public static ArrayList<NetworkMeasurement> mNetworkMeasurementsRunning = new ArrayList<>();
-    public static ArrayList<NetworkMeasurement> mNetworkMeasurementsFinished = new ArrayList<>();
+    //public static ArrayList<NetworkMeasurement> mNetworkMeasurementsRunning = new ArrayList<>();
+    //public static ArrayList<NetworkMeasurement> mNetworkMeasurementsFinished = new ArrayList<>();
```
bassosimone: @lorenzoPrimi if these two arrays are not needed anymore, I'd remove them rather than just commenting 'em out

app/src/main/java/org/openobservatory/netprobe/data/TestStorage.java
```diff
@@ -0,0 +1,96 @@
...
+        settings = context.getSharedPreferences(PREFS_NAME,Context.MODE_PRIVATE);
+        if (settings.contains(TESTS)) {
+            String jsonTests = settings.getString(TESTS, null);
+            Gson gson = new Gson();
+            NetworkMeasurement[] favoriteItems = gson.fromJson(jsonTests,NetworkMeasurement[].class);
+            tests = Arrays.asList(favoriteItems);
+            tests = new ArrayList(tests);
+        } else
+            return new ArrayList();
+        return (ArrayList) tests;
```
bassosimone: Usually I see the cast from ArrayList to List, not the other way round. Why here you are casting from the general class to the particular class? I thought it was more common to return the general class.

app/src/main/java/org/openobservatory/netprobe/data/TestStorage.java
```diff
@@ -0,0 +1,96 @@
...
+        List tests;
+        settings = context.getSharedPreferences(PREFS_NAME,Context.MODE_PRIVATE);
+        if (settings.contains(TESTS)) {
+            String jsonTests = settings.getString(TESTS, null);
+            Gson gson = new Gson();
+            NetworkMeasurement[] favoriteItems = gson.fromJson(jsonTests,NetworkMeasurement[].class);
+            tests = Arrays.asList(favoriteItems);
+            tests = new ArrayList(tests);
+        } else
+            return new ArrayList();
```
bassosimone: You can perhaps refactor this block as

```java
if (!settings.contains(TESTS)) {
    return new ArrayList();
}
// remainder of the code here
```

app/src/main/java/org/openobservatory/netprobe/data/TestStorage.java
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    public void setCompleted(Context context, NetworkMeasurement test) {
+        List tests = loadTests(context);
+        if (tests != null){
+            for(int i = 0; i < tests.size(); i++) {
+                NetworkMeasurement n = (NetworkMeasurement)tests.get(i);
+                if (n.test_id == test.test_id) {
+                    n.completed = true;
+                    tests.set(i, n);
```
bassosimone: Is it correct to say that here you're looking for the fist occurrence only? If so, you should break out of the loop as soon as you have find the fist occurrence, right?

app/src/main/java/org/openobservatory/netprobe/data/TestStorage.java
```diff
@@ -0,0 +1,96 @@
...
+        storeTests(context, tests);
+    }
+
+    public void removeTest(Context context, NetworkMeasurement test) {
+        List tests = loadTests(context);
+        if (tests != null){
+            for(int i = 0; i < tests.size(); i++) {
+                NetworkMeasurement n = (NetworkMeasurement)tests.get(i);
+                if (n.test_id == test.test_id) {
+                    tests.remove(i);
```
bassosimone: Same here: unless more than one test can have the same id, you should break out of the loop after you have found the matching occurrence.

app/src/main/java/org/openobservatory/netprobe/utils/Alert.java
bassosimone: @lorenzoPrimi have you checked whether this could be done using `Gson`? I'd rather merge a pull request containing non custom code for increased maintainability.

app/src/main/java/org/openobservatory/netprobe/view/ListImageButton.java
```diff
@@ -0,0 +1,43 @@
...
+
+import android.content.Context;
+import android.util.AttributeSet;
+import android.view.View;
+import android.widget.ImageButton;
+
+
+/**
+ * Un bottone definito per essere utilizzato in una {@link ListView}.
+ */
```
bassosimone: It would be great if comments could be in English

bassosimone:
Good work, @lorenzoPrimi, 👍 !

I did another pass of review of the pull request and pointed to some little issues that I'd like to be addressed before merging the pull request!

app/src/main/java/org/openobservatory/netprobe/activity/MainActivity.java
lorenzoPrimi: Don't think so, it's still a TODO

Makefile
lorenzoPrimi: Where is my mind? Ops I mean where is the push? :)

app/src/main/java/org/openobservatory/netprobe/utils/Alert.java
lorenzoPrimi: We don't need it anymore, that code is going to be removed with the new ooni-wui

app/src/main/java/org/openobservatory/netprobe/data/TestData.java
lorenzoPrimi: agreed

app/src/main/java/org/openobservatory/netprobe/data/TestStorage.java
lorenzoPrimi: Actually I don't remember, maybe it works both way, need to test it

app/src/main/java/org/openobservatory/netprobe/data/TestStorage.java
lorenzoPrimi: Agreed

Makefile
bassosimone: > "way down in the water... see it swimming"

I am uploading the newly compiled binaries right now; yesterday apparently the process was interrupted

app/src/main/java/org/openobservatory/netprobe/activity/MainActivity.java
bassosimone: Is it possible to add a patch addressing that in the context of this pull request? I think it would be a good thing to do that, so when we merge we merge a branch that works as it ought to work

app/src/main/java/org/openobservatory/netprobe/data/TestStorage.java
bassosimone: Thank you!

app/src/main/java/org/openobservatory/netprobe/data/TestStorage.java
bassosimone: Addressed!

app/src/main/java/org/openobservatory/netprobe/utils/Alert.java
bassosimone: Okay, great!

bassosimone:
@lorenzoPrimi I have committed dd1271d1 to use the latest version of MK that I finally rebuilt and started testing the app in the emulator. As soon as I start a test there is a crash because the SSL certificate bundle is not present. I think we should add one as we do for iOS.

lorenzoPrimi:
@bassosimone there is no way to do that until I have the set_options function

bassosimone:
Okay, @lorenzoPrimi, I've added a fix such that reports are submitted, such that we can move forward, thank you for working on this, great work! Now I see how it's becoming more and more pressing to have an API that allows us to do exactly the same we do in C++.

PR #12, opened 2016-08-20
Use SWIG API and MK 0.3.0

bassosimone:
@lorenzoPrimi to test this you need to recompile yourself the `-android` repository and copy the files in the rootdir of the repository and then run `make unpack`. In my experience, it's better to close AndroidStudio, clean up everything in the repo (using `git clean -dfx`), run `make unpack`, and then start AndroidStudio again.


app/src/main/java/org/openobservatory/netprobe/data/TestData.java
```diff
@@ -167,4 +167,8 @@ public void notifyObservers(Object type) {
         }
         return servers;
     }
+
+    public static String boolToString(Boolean b) {
+        return b ? "1" : "0";
+    }
```
bassosimone: @lorenzoPrimi, yes, this looks good to me :-)

bassosimone:
Okay, @lorenzoPrimi, with 383d841 committed, I think all the pieces to use MK v0.3.0 are there, except that one needs to modify the way in which tests are executed to be like TcpConnect rather than using the style used for other tests. Also, now this is based v0.3.0 so there is a NDT test working just like other OONI tests and there is also web_connectivity!

PR #68, opened 2017-03-03
Release/1.1.3

lorenzoPrimi:


hellais:
The ChangeLog need to be updated with changes introduced in this version.

app/build.gradle
```diff
@@ -14,8 +14,8 @@ android {
         applicationId "org.openobservatory.ooniprobe"
         minSdkVersion 9
         targetSdkVersion 25
-        versionCode 4
-        versionName "1.1.2"
+        versionCode 5
+        versionName "1.1.3-dev"
```
hellais: It's good that this is while we are developing `-dev`. Remember to bump this up to `1.1.3` when merging.

app/src/main/java/org/openobservatory/ooniprobe/activity/MainActivity.java
```diff
@@ -121,12 +118,13 @@ public void onDrawerStateChanged(int newState){
             selectItem(0);
         }
 
-        checkInformedConsent();
-    }
+        SharedPreferences preferences = PreferenceManager.getDefaultSharedPreferences(this);
+        if (!preferences.getBoolean("cleanup_unused_files", false)) {
+            TestStorage.removeUnusedFiles(this);
+            PreferenceManager.getDefaultSharedPreferences(this).edit().putBoolean("cleanup_unused_files", true).apply();
+        }
```
hellais: I imagine this is the cleanup code for the issue with files not really being deleted when you deleted them on Android?

If so the logic is to run it only once for the first time, correct? What is the plan for the future, like are we from now on always going to be checking for this?

app/build.gradle
lorenzoPrimi: Yes, this will be done before pushing it to the stores

app/src/main/java/org/openobservatory/ooniprobe/activity/MainActivity.java
lorenzoPrimi: We can either keep it or remove it after x months where x > 6

hellais:
We should also ensure to update the test-list to the latest version.

lorenzoPrimi:
Maybe we can write a script to auto update the list for both projects?

hellais:
Before merging we should:

* [ ] Update to the currently global testing list (see: https://github.com/OpenObservatory/ooni-resources/releases/download/9/citizenlab-test-lists.global.csv)

* [ ] Bump version number to not end with `-dev` (this should be bumped when all the outstanding issues identified in PR review have been addressed)

lorenzoPrimi:
For the first point, can we make a script in the root folder that overrides app/src/main/res/raw/global.txt ?

hellais:
> For the first point, can we make a script in the root folder that overrides app/src/main/res/raw/global.txt ?

I added a script to both repos to do this.

lorenzoPrimi:
Great, I guess we are ready to merge and push both apps

PR #78, opened 2017-03-21
NetworkMeasurement.java: initialize all attributes

bassosimone:
Make sure all attributes are initialized, so to ensure that there
are no surprises if the constructor is not called.

It might be that the diff would be correct in C++ and correct and
overkill in Java, but what mattes is that it fixes #76.

Now, there is another pending diff to make sure we have completely
addressed all the issues exposed during #76.

Issue #76, opened 2017-03-20
Crash when visualising results with tests in progress


bassosimone:
On Android emulator, architecture i386, API 24, with 124f32c570 as the tip, you can sometimes provoke a crash by following this procedure (it occurred to me twice, once with another emulator and I am trying to understand how to make this more easily reproducible):

- run all tests concurrently
- as soon as http-invalid-request-line terminates, go to results view

When you're lucky (well, or unlucky, I guess that depends on your objective function), the app will crash like this:

```
03-20 16:00:49.860 2252-2252/org.openobservatory.ooniprobe E/AndroidRuntime: FATAL EXCEPTION: main
 Process: org.openobservatory.ooniprobe, PID: 2252
 java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.Boolean.booleanValue()' on a null object reference
     at org.openobservatory.ooniprobe.data.TestStorage.loadTestsReverse(TestStorage.java:69)
     at org.openobservatory.ooniprobe.fragment.PastTestsFragment.updateList(PastTestsFragment.java:94)
     at org.openobservatory.ooniprobe.fragment.PastTestsFragment.onResume(PastTestsFragment.java:51)
     at android.support.v4.app.Fragment.performResume(Fragment.java:2235)
     at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1346)
     at android.support.v4.app.FragmentManagerImpl.moveFragmentToExpectedState(FragmentManager.java:1528)
     at android.support.v4.app.FragmentManagerImpl.moveToState(FragmentManager.java:1595)
     at android.support.v4.app.BackStackRecord.executeOps(BackStackRecord.java:757)
     at android.support.v4.app.FragmentManagerImpl.executeOps(FragmentManager.java:2355)
     at android.support.v4.app.FragmentManagerImpl.executeOpsTogether(FragmentManager.java:2146)
     at android.support.v4.app.FragmentManagerImpl.optimizeAndExecuteOps(FragmentManager.java:2098)
     at android.support.v4.app.FragmentManagerImpl.execPendingActions(FragmentManager.java:2008)
     at android.support.v4.app.FragmentManagerImpl$1.run(FragmentManager.java:710)
     at android.os.Handler.handleCallback(Handler.java:751)
     at android.os.Handler.dispatchMessage(Handler.java:95)
     at android.os.Looper.loop(Looper.java:154)
[...]

PR #79, opened 2017-03-21
MainActivity.java: more robust resource copying

bassosimone:
Rather than relying on a shared preference, directly check whether
the file is there or not and, in the latter case, copy it.

Observed during #76. Depends on #78.

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/activity/MainActivity.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/activity/MainActivity.java
@@ -30,6 +30,7 @@ import android.widget.ListView;
 import android.widget.TextView;
 import android.widget.Toast;
 
+import java.io.FileNotFoundException;
 import java.io.FileOutputStream;
 import java.io.InputStream;
 import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


bassosimone:
@AntonioLangiu, could you review #78 and #79?

AntonioLangiu:
Yess :+1:

app/src/main/java/org/openobservatory/ooniprobe/activity/MainActivity.java
```diff
@@ -272,18 +273,24 @@ public void update(Observable observable, Object data) {
...
+        boolean exists = false;
+        try {
+            openFileInput(filename);
+            exists = true;
+        } catch (FileNotFoundException exc) {
+            /* FALLTHROUGH */
+        }
+        if (exists) {
+            return;
+        }
```
AntonioLangiu: This part can be written in less lines, like:

    try {
        openFileInput(filename);
        return;
    } catch (FileNotFoundException exc) {
        /* FALLTHROUGH */
    }

but in the way you wrote it increases readability.

bassosimone: Yep, initially it was written as you say, then I erred on the side of readability!

bassosimone:
@AntonioLangiu thank you for looking into this, merging!

PR #81, opened 2017-03-25
Fixes image redirect on README

frankcash:
When clicking the header image of the README it redirects to the actual image hosted on github.  It now redirects to the OONI website.


frankcash:
Also adds code styling to a src path

hellais:
Thanks for this!

PR #82, opened 2017-04-10
Download AAR from jcenter and use it

bassosimone:
<strike>This implementation also validates the AAR using gpg. For this reason we don' use gradle. It seems there is no satisfying, automated way to verify digital signature of dependencies.</strike>

Use gradle. Discussed with @hellais.


bassosimone:
I have discussed this with @hellais and we'll follow here the same route followed in https://github.com/measurement-kit/android-example/pull/11

build.gradle
```diff
@@ -7,14 +7,17 @@ buildscript {
...
         classpath 'io.fabric.tools:gradle:1.+'
     }
 }
 
 allprojects {
     repositories {
         jcenter()
+        flatDir {
+            dirs 'libs'
+        }
```
bassosimone: This ensures that we can either fetch from gradle or put directly in here the `aar` thus skipping fetching from gradle.

scripts/check-version-consistency.sh
```diff
@@ -0,0 +1,23 @@
...
+# Convenience script to make sure the MK AAR in `libs` matches the
+# one listed in `app/build.gradle`.
+#
+
+set -e
+versions=`ls libs/android-libs-*.aar`
+count=0
+for version in $versions; do
+    count=$((count + 1))
+done
```
bassosimone: This can probably be simplified by running `ls $stuff | wc -l`

scripts/fetch-aar.sh
```diff
@@ -0,0 +1,25 @@
...
+# Convenience script to fetch the AAR and validate its digital signature
+# when you don't want to download it automatically using gradle.
+#
+
+set -e
+if [ $# -ne 1 ]; then
+    echo "usage: $0 version" 1>&2
+    exit 1
+fi
+rm -rf libs/*.aar
```
bassosimone: This does not seem necessary

bassosimone: We should probably delete `android-libs-$version.aar` only, so wget does not clobber it. Or we can configure wget not to create backup files and directly overwrite them if they're already there (<- this seems more idempotent)

scripts/fetch-aar.sh
```diff
@@ -0,0 +1,25 @@
...
+version=$1
+file=android-libs-$version.aar
+url="https://dl.bintray.com/measurement-kit/android/org/openobservatory/measurement_kit/android-libs/$version/$file"
+wget $url $url.asc
+gpg --keyserver keyserver.ubuntu.com                                           \
+  --recv-keys 738877AA6C829F26A431C5F480B691277733D95B
```
bassosimone: I should probably add a comment specifying that this is my PGP key

PR #85, opened 2017-05-04
Norwegian Bokmål translation.

comradekingu:


hellais:
Would it be possible to add these to transifex? That way they would also work for the iOS app as well. Also there are more strings to translate that are part of the GUI that are shared by both the iOS and Android apps in the web UI and those are in the web UI.

comradekingu:
@hellais I have translated it on TX now.

hellais:
@comradekingu amazing! Thanks so much 👍

PR #92, opened 2017-07-05
Update to MK v0.7.0-alpha.2-2

bassosimone:
This diff updates the Android app to use MK v0.7.0-alpha.2-2. This version of MK includes code for performing orchestration. [See example in android-libs](https://github.com/measurement-kit/android-libs/blob/feature/orchestration/example/java/OrchestrateExample.java).


app/src/main/java/org/openobservatory/ooniprobe/model/PortolanTests.java
```diff
@@ -1,47 +0,0 @@
...
-                Log.v(TAG, "other -> break");
-                break;
-            }
-        }
-        PortolanSyncApi.closeProber(prober);
-        Log.v(TAG, "runTraceroute... done");
-    }
-
-    private static final String TAG = "portolan-tests";
-}
```
lorenzoPrimi: Is this class totally useless now?

bassosimone: The JNI code for it has been disabled for quite some time. Since it's not priority, I will write updated code when we need it again. Until now, better not to keep around a class that, if invoked, leads to a crash because specific JNI functions that it expects are missing.

PR #103, opened 2017-09-07
Feature/uri scheme

lorenzoPrimi:


app/src/main/AndroidManifest.xml
```diff
@@ -23,6 +23,13 @@
                 <action android:name="android.intent.action.MAIN" />
 
                 <category android:name="android.intent.category.LAUNCHER" />
+
+                <action android:name="android.intent.action.VIEW" />
+
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+
+                <data android:scheme="ooni" />
```
hellais: The intent filter for `run.ooni.io` is missing. See: https://developer.android.com/training/app-links/deep-linking.html

hellais: It should look something like:

```
    <intent-filter android:label="@string/nettest">
        <data android:scheme="http"
              android:host="run.ooni.io"
              android:pathPrefix="/nettest" />
    </intent-filter>
```

hellais: Also this intent filter for the `ooni` scheme is too broad. It should be restricted to `ooni://nettest` like so:

```
    <intent-filter android:label="@string/nettest">
        <data android:scheme="ooni"
              android:host="nettest" />
    </intent-filter>
```

lorenzoPrimi: I didn't know we agreed on the http scheme other the ooni:// one, next time update specs so I don't lose what we write on the channel

hellais: This is about adding support for universal links. It's the thing we discussed for hours and days :P

lorenzoPrimi: The  android:host="nettest" is not needed as I check if the url is supported in the app, but I added it anyway.

app/src/main/AndroidManifest.xml
```diff
@@ -74,15 +73,46 @@
...
+                <data android:scheme="https"
+                    android:host="run.ooni.io"
+                    android:pathPrefix="/nettest" />
+            </intent-filter>
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
+                <action android:name="org.openobservatory.ooniprobe.nettest" />
+                <data android:scheme="http"
```
bassosimone: Do we want to support `http://run.ooni.io` links? Cc: @hellais

app/src/main/AndroidManifest.xml
```diff
@@ -74,15 +73,46 @@
...
+            android:name=".activity.NotificationDialog"
             android:theme="@style/Theme.Transparent" />
+        <activity
+            android:name=".activity.RunTestActivity"
+            android:label="@string/app_name"
+            android:launchMode="singleInstance">
+            <intent-filter>
+                <action android:name="android.intent.action.VIEW" />
+                <category android:name="android.intent.category.DEFAULT" />
+                <category android:name="android.intent.category.BROWSABLE" />
```
bassosimone: So, here I understand that `BROWSABLE` category is the one set when the user clicks a link in the browser and you have registered as handler for such link, correct? I wonder in which case we receive intents with no category at all is set or have you added that category just as a catch-all and just in case?

app/src/main/AndroidManifest.xml
```diff
@@ -74,15 +73,46 @@
...
             </intent-filter>
         </activity>
-        <activity android:name=".activity.NotificationDialog"
+        <activity
+            android:name=".activity.NotificationDialog"
             android:theme="@style/Theme.Transparent" />
+        <activity
+            android:name=".activity.RunTestActivity"
+            android:label="@string/app_name"
+            android:launchMode="singleInstance">
```
bassosimone: @lorenzoPrimi do I remember it correctly that this `singleInstance` here is what saved us a lot of bugs such that the activity was recycled and so we could not open the app again for the same link? If so, maybe let's document this, because otherwise maybe in the future we don't remember why we set that flag; what do you think?

app/src/main/java/org/openobservatory/ooniprobe/activity/BrowserActivity.java
```diff
@@ -148,6 +150,10 @@ public void onClick(View view) {
         }
     }
 
+    @Override
+    protected void attachBaseContext(Context newBase) {
+        super.attachBaseContext(CalligraphyContextWrapper.wrap(newBase));
+    }
```
bassosimone: Okay, here you've basically added support for calligraphy inside the browser. I believe small changes like this can be submitted as quick-review pull requests that we can merge. Then, as I was saying on the iOS repository, we can use github's feature that we can merge master back into the PR and we get these small fixes in the branch anyway.

app/src/main/java/org/openobservatory/ooniprobe/activity/MainActivity.java
```diff
@@ -132,17 +137,12 @@ public void onDrawerStateChanged(int newState) {
...
-        // XXX: This is probably not correct: we would like to send
-        // info to the orchestrator only when the network or any other
-        // orchestrator parameter like country code changed.
-        //TODO remove this code in 1.1.6
-        String token = FirebaseInstanceId.getInstance().getToken();
-        if (token != null) {
-            NotificationService ns = NotificationService.getInstance(this);
-            ns.setDevice_token(token);
-            ns.sendRegistrationToServer();
-        }
```
bassosimone: FTR, we verified with @lorenzoPrimi that this was causing two back-to-back orchestration requests.

app/src/main/java/org/openobservatory/ooniprobe/activity/MainActivity.java
```diff
@@ -66,6 +70,7 @@
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
+
         setContentView(R.layout.activity_main);
         checkResources();
         TestData.getInstance(this, this).addObserver(this);
```
bassosimone: My understanding of the [Activity lifecycle](https://developer.android.com/guide/components/activities/activity-lifecycle.html) is that `onResume` is called after `onCreate`; hence you don't need to `addObserver` this because you already do the same below inside of `onResume`.

app/src/main/java/org/openobservatory/ooniprobe/activity/MainActivity.java
```diff
@@ -132,17 +137,12 @@ public void onDrawerStateChanged(int newState) {
...
-        String token = FirebaseInstanceId.getInstance().getToken();
-        if (token != null) {
-            NotificationService ns = NotificationService.getInstance(this);
-            ns.setDevice_token(token);
-            ns.sendRegistrationToServer();
-        }
+    @Override
+    public void onResume() {
+        super.onResume();
+        TestData.getInstance(this, this).addObserver(this);
```
bassosimone: So, I've read the code of `getInstance` and `addObserver`. Now, I also get why this is called `TestData` :-).

What is not totally clear to me is the logic of `getInstance`: can you explain it? (I believe this could also be useful knowledge to me for refactoring the tests to use intents so to fully decouple running tests and view.)

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
```diff
@@ -0,0 +1,236 @@
...
+    private static String test_name;
+
+    @Override
+    protected void onCreate(Bundle savedInstanceState) {
+        super.onCreate(savedInstanceState);
+        setContentView(R.layout.activity_run_test);
+
+        // Get the intent that started this activity
+        Intent intent = getIntent();
+        Uri data = intent.getData();
```
bassosimone: I think this variable here is unused

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
```diff
@@ -0,0 +1,236 @@
...
+        if (Intent.ACTION_VIEW.equals(intent.getAction())) {
+            Uri uri = intent.getData();
+            String mv = uri.getQueryParameter("mv");
```
bassosimone: handy!

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
```diff
@@ -0,0 +1,236 @@
...
+                    String tn = uri.getQueryParameter("tn");
+                    String ta = uri.getQueryParameter("ta");
+                    String td = uri.getQueryParameter("td");
+                    String test = NetworkMeasurement.getTestName(this, tn);
+                    if (test.length() > 0){
+                        test_name = tn;
+                        configureScreen(td, ta);
+                    }
+                    else {
+                        Alert.alertDialog(this, getString(R.string.invalid_parameter), getString(R.string.test_name) +  " : " + tn);
```
bassosimone: Does this mean that we don't redirect the user to the store if the version is too small?

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
```diff
@@ -0,0 +1,236 @@
...
+                            }
+                        }
+                    });
+                }
+            }
+        }
+    }
+
+    public void configureScreen(String td, String ta){
+        TestData.getInstance(this, this).addObserver(this);
```
bassosimone: My understanding:

- when the app is started by the main activity, the `context` of `TestData` is the `MainActivity` and the `activity` of `TestData` is the main activity;

- when the app is started by this activity, `context` and `activity` are this activity.

I still haven't understood what happens when this activity exits. In the case in which the app also quits, then probably okay. Whereas, if the main activity is somehow started, wouldn't this mean that we're leaking this activity because it will never be freed since there is a reference loop between `TestData` and this activity and, moreover, `TestData` is referenced by a live object, i.e. the `MainActivity`?

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
```diff
@@ -0,0 +1,236 @@
...
+        ArrayList<String> listItems = new ArrayList<>();
+        final ArrayList<String> urlItems = new ArrayList<>();
+        if (ta != null){
+            try {
+                JSONObject taJson = new JSONObject(ta);
+                JSONArray urls = taJson.getJSONArray("urls");
+                for (int i = 0; i < urls.length(); i++)
+                    urlItems.add(urls.getString(i));
+            } catch (JSONException e) {
+                e.printStackTrace();
```
bassosimone: Do we want to continue processing in this case?

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
```diff
@@ -0,0 +1,236 @@
...
+            mUrlListAdapter = new UrlListAdapter(this, listItems, true);
+
+        if (listItems.size() == 0)
+            urls.setVisibility(View.INVISIBLE);
+
+        runButton = (AppCompatButton) findViewById(R.id.run_test_button);
+        runButton.setOnClickListener(
+                new ImageButton.OnClickListener() {
+                    public void onClick(View v) {
+                        TestData.doNetworkMeasurements(getApplicationContext(), test_name, urlItems);
```
bassosimone: I am a bit confused by the fact that sometimes you call `TestData`'s singleton and some other times you don't. This makes it more complex for me to reason about the lifecycle of the singleton.

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
```diff
@@ -0,0 +1,236 @@
...
+                        Intent MainActivityIntent = new Intent(RunTestActivity.this, MainActivity.class);
+                        MainActivityIntent.setFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
+                        MainActivityIntent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
```
bassosimone: It would be helpful to explain why the activity is being launched with these flags

app/src/main/java/org/openobservatory/ooniprobe/fragment/SettingsFragment.java
```diff
@@ -304,6 +304,16 @@ public void onClick(DialogInterface dialog, int which) {
...
+        builder.setNeutralButton(R.string.set_default, new DialogInterface.OnClickListener() {
+            @Override
+            public void onClick(DialogInterface dialog, int which) {
+                SharedPreferences.Editor editor = preferences.edit();
+                editor.putString("collector_address", OONITests.COLLECTOR_ADDRESS);
+                editor.commit();
+                TextView collector_address = (TextView)mActivity.findViewById(R.id.collector_address_subText);
+                collector_address.setText(OONITests.COLLECTOR_ADDRESS);
+            }
+        });
```
bassosimone: Didn't we already support this?

app/src/main/java/org/openobservatory/ooniprobe/model/NetworkMeasurement.java
```diff
@@ -28,6 +29,7 @@ public NetworkMeasurement(String name) {
         this.anomaly = 0;
     }
 
+    @NonNull
```
bassosimone: Here you are basically hinting (or asserting?) that the return value is `non-null`?

app/src/main/java/org/openobservatory/ooniprobe/model/NetworkMeasurement.java
```diff
@@ -40,11 +42,13 @@ public static String getTestName(Context context, String name) {
                 return context.getString(R.string.http_invalid_request_line);
             case OONITests.HTTP_HEADER_FIELD_MANIPULATION:
                 return context.getString(R.string.http_header_field_manipulation);
+            case OONITests.NDT:
+                return context.getString(R.string.ndt);
             case OONITests.NDT_TEST:
                 return context.getString(R.string.ndt);
```
bassosimone: Uhm, shouldn't you delete the case where `NDT_TEST` is used?

app/src/main/java/org/openobservatory/ooniprobe/model/NetworkMeasurement.java
```diff
@@ -59,6 +63,8 @@ public static String getTestDescr(Context context, String name) {
                 return context.getString(R.string.http_header_field_manipulation_desc);
             case OONITests.NDT_TEST:
                 return context.getString(R.string.ndt_desc);
+            case OONITests.NDT:
+                return context.getString(R.string.ndt_desc);
             case OONITests.DASH:
```
bassosimone: Ditto

app/src/main/java/org/openobservatory/ooniprobe/model/NetworkMeasurement.java
```diff
@@ -96,6 +102,13 @@ else if (anomaly == 1)
                     return R.drawable.ndt_warning;
                 else
                     return R.drawable.ndt_no;
+            case OONITests.NDT:
+                if (anomaly == 0)
+                    return R.drawable.ndt;
+                else if (anomaly == 1)
+                    return R.drawable.ndt_warning;
+                else
+                    return R.drawable.ndt_no;
```
bassosimone: Ditto

app/src/main/java/org/openobservatory/ooniprobe/model/OONITests.java
```diff
@@ -11,6 +11,7 @@
     public static final String HTTP_HEADER_FIELD_MANIPULATION = "http_header_field_manipulation";
     public static final String TCP_CONNECT = "tcp_connect";
     public static final String WEB_CONNECTIVITY = "web_connectivity";
+    public static final String NDT = "ndt";
     public static final String NDT_TEST = "ndt_test";
```
bassosimone: Ah, uhm, now I see this is defined both as `NDT` and as `NDT_TEST`: why?

app/src/main/java/org/openobservatory/ooniprobe/utils/AlarmReceiver.java
```diff
@@ -16,9 +16,9 @@ public void onReceive(Context context, Intent intent) {
         //System.out.println(DEBUG_TAG);
         NotificationHandler.sendNotification(context, context.getString(R.string.local_notifications_text));
         TestData.getInstance(context, null);
-        TestData.doNetworkMeasurements(context, "web_connectivity");
-        TestData.doNetworkMeasurements(context, "http_invalid_request_line");
-        TestData.doNetworkMeasurements(context, "ndt_test");
+        TestData.doNetworkMeasurements(context, "web_connectivity", null);
+        TestData.doNetworkMeasurements(context, "http_invalid_request_line", null);
+        TestData.doNetworkMeasurements(context, "ndt_test", null);
```
bassosimone: @hellais do we want to run more tests here?

app/src/main/java/org/openobservatory/ooniprobe/utils/NotificationService.java
```diff
@@ -80,13 +80,16 @@ public void sendRegistrationToServer() {
...
+        /*
         System.out.println("probe_cc: " + geoip_country_path);
         System.out.println("probe_asn: " + geoip_asn_path);
         System.out.println("platform: " + platform);
         System.out.println("software_name: " + software_name);
         System.out.println("software_version: " + software_version);
         System.out.println("supported_tests: " + supported_tests);
         System.out.println("token: " + device_token);
+        */
+        if (device_token == null) return;
```
bassosimone: Maybe we can log the fact that the device is `null` so we know when we're debugging?

ChangeLog.md
```diff
@@ -1,3 +1,15 @@
+# ooniprobe-android 1.2.0-rc.2 [2017-09-10]
+
+New features:
+
+* Added support for running tests with custom URI scheme
+
+* Fixed bug on HTTP Header Field Manipulation result list
+
+measurement-kit version: 0.7.3-3
```
bassosimone: This should be `v0.7.2-1`

bassosimone: Or maybe was `v0.7.3-3`.

app/src/main/AndroidManifest.xml
bassosimone: As reference I'm reading [the link you guys circulated on slack](https://inthecheesefactory.com/blog/understand-android-activity-launchmode/en).

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
bassosimone: Also, I was wondering: the `MainActivity` is by default created with what flags? Does it matter that here we're launching it with non-default flags?

app/src/main/AndroidManifest.xml
lorenzoPrimi: Usually is a bad practice to comment the xml files in Android, I haven't seen it anywhere

app/src/main/java/org/openobservatory/ooniprobe/activity/BrowserActivity.java
lorenzoPrimi: This should be attached to every class to use the same font. Now the browser didn't need it because we have no labels, but better adding it

app/src/main/java/org/openobservatory/ooniprobe/activity/MainActivity.java
lorenzoPrimi: Didn't noticed it was duplicated

app/src/main/java/org/openobservatory/ooniprobe/activity/MainActivity.java
lorenzoPrimi: I need which class is "active" to manage the Observables

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
lorenzoPrimi: What else do you want to do? You just don't add urls

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
lorenzoPrimi: TestData.doNetworkMeasurements is always called in the same way. Because it's already been created

app/src/main/java/org/openobservatory/ooniprobe/fragment/SettingsFragment.java
lorenzoPrimi: No, not autocomplete

app/src/main/java/org/openobservatory/ooniprobe/model/NetworkMeasurement.java
lorenzoPrimi: This is needed because of old apps have ndt_test instead of ndt string

app/src/main/java/org/openobservatory/ooniprobe/model/NetworkMeasurement.java
lorenzoPrimi: same

app/src/main/java/org/openobservatory/ooniprobe/model/OONITests.java
lorenzoPrimi: same as above

app/src/main/java/org/openobservatory/ooniprobe/utils/AlarmReceiver.java
lorenzoPrimi: Good question

app/src/main/java/org/openobservatory/ooniprobe/utils/NotificationService.java
lorenzoPrimi: This scenario should happen only if user deacivated the notification.

ChangeLog.md
```diff
@@ -1,3 +1,39 @@
+# ooniprobe-android 1.2.0 [2017-09-xx]
```
bassosimone: @lorenzoPrimi remember to change the release date

app/build.gradle
```diff
@@ -14,8 +14,8 @@ android {
         applicationId "org.openobservatory.ooniprobe"
         minSdkVersion 14
         targetSdkVersion 25
-        versionCode 11
-        versionName "1.1.5"
+        versionCode 12
+        versionName "1.2.0"
```
bassosimone: Here on Android, where you can do whatever you want with `versionName`, I suggest to add the `-rc` suffix and only leave it when we're about to release.

app/build.gradle
```diff
@@ -37,7 +37,7 @@ dependencies {
         transitive = true;
     }
     //compile 'com.android.support:customtabs:25.3.1'
-    compile 'org.openobservatory.measurement_kit:android-libs:0.7.0-1'
+    compile 'org.openobservatory.measurement_kit:android-libs:0.7.2-1'
     compile 'com.android.support:appcompat-v7:25.3.1'
     compile 'com.android.support:recyclerview-v7:25.3.1'
     compile 'uk.co.chrisjenx:calligraphy:2.2.0'
```
bassosimone: @lorenzoPrimi: can you also make sure we're not using any alpha or beta software in this file, because having alpha or beta in here breaks the `f-droid` build. Here it seems you only need to bump `constraint-layout`.

app/src/main/AndroidManifest.xml
bassosimone: Fair enough. We should still document somewhere _why_ we start activities the way we do.

app/src/main/AndroidManifest.xml
bassosimone: ^^^ @lorenzoPrimi

app/src/main/AndroidManifest.xml
bassosimone: Ping @hellais: do we want to support non encrypted `run.ooni.io` links? I guess the possible reason why we want to do that is that one _may_ write the link without `https`. Do we really want this or do we want to enforce `https`?

app/src/main/java/org/openobservatory/ooniprobe/activity/BrowserActivity.java
bassosimone: Yes, I agree. Thanks for explaining. (As I said, I think things like this can be pushed as separate hotfix PRs.)

app/src/main/java/org/openobservatory/ooniprobe/activity/MainActivity.java
bassosimone: I understand. I think that's not super obvious from the API. As I said, I find the fact that sometimes a singleton is used and sometimes static methods a bit strange.

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
bassosimone: ^^^ @lorenzoPrimi?

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
bassosimone: I don't know, perhaps do not run the tests because the JSON is malformed?

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
bassosimone: What I mean is that sometimes you access the singleton of `TestData` (for example for setting the observers) and in other cases like this instead you call a static method. This does not make obvious what is the singleton used for and whether the singleton is used or not when you call the static method, unless one reads the code.

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
bassosimone: ^^^ @lorenzoPrimi

app/src/main/java/org/openobservatory/ooniprobe/data/TestData.java
```diff
@@ -410,39 +417,18 @@ public static void setAnomaly_hhfm(String entry, NetworkMeasurement test){
...
-                if (test_keys.has("header_name_capitalization"))
-                    header_name_capitalization = tampering.getBoolean("header_name_capitalization");
-                if (test_keys.has("request_line_capitalization"))
-                    request_line_capitalization = tampering.getBoolean("request_line_capitalization");
-                if (test_keys.has("total"))
-                    total = tampering.getBoolean("total");
-
-                if (header_field_name || header_field_number || header_field_value || header_name_capitalization || request_line_capitalization || total)
-                    anomaly = 2;
+                String keys[] = {"header_field_name","header_field_number","header_field_value", "header_name_capitalization", "request_line_capitalization", "total"};
```
bassosimone: @lorenzoPrimi can you wrap this very long line?

app/src/main/java/org/openobservatory/ooniprobe/data/TestData.java
```diff
@@ -410,39 +417,18 @@ public static void setAnomaly_hhfm(String entry, NetworkMeasurement test){
...
+                for (String key: keys)
+                {
+                    if (tampering.has(key))
+                        if (tampering.getBoolean(key))
+                            anomaly = 2;
```
bassosimone: I'd rather avoid using bare numbers. Can't we define static variables (or the Java equivalent of `#define`s) such that we can basically understand better what each "define" means? (`ANOMALY` is more clear than `2`)

app/src/main/java/org/openobservatory/ooniprobe/fragment/SettingsFragment.java
bassosimone: Ah, I understand!

app/src/main/java/org/openobservatory/ooniprobe/model/NetworkMeasurement.java
bassosimone: Why is it needed? This is a new app

app/src/main/java/org/openobservatory/ooniprobe/model/NetworkMeasurement.java
```diff
@@ -116,7 +129,7 @@ public static int getTestImageBig(String name) {
                 return R.drawable.http_invalid_request_line_big;
             case OONITests.HTTP_HEADER_FIELD_MANIPULATION:
                 return R.drawable.http_header_field_manipulation_big;
-            case OONITests.NDT_TEST:
+            case OONITests.NDT:
```
bassosimone: Here, for example, why are you `s/NDT_TEST/NDT/` and not adding another case?

app/src/main/java/org/openobservatory/ooniprobe/utils/AlarmReceiver.java
bassosimone: Ping @hellais

app/src/main/java/org/openobservatory/ooniprobe/utils/NotificationService.java
bassosimone: Okay, maybe a comment should then be useful to clarify _when_ this could happen

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
lorenzoPrimi: Whenever you enter MainActivity I re-set the activity to that. Either MainActivity or RunTestActivity will be running

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
lorenzoPrimi: the parameter name test name is already validated, do you think we should abort all because the urls are malformed?

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
lorenzoPrimi: So I guess I should change everything to singleton

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
lorenzoPrimi: Don't know, just copied from Android manual.

app/src/main/AndroidManifest.xml
lorenzoPrimi: Don't know, just copied from Android manual.

app/src/main/java/org/openobservatory/ooniprobe/model/NetworkMeasurement.java
lorenzoPrimi: Don't know, just copied from a working example

app/src/main/java/org/openobservatory/ooniprobe/model/NetworkMeasurement.java
lorenzoPrimi: Because the stored tests on the device can have the old test name

app/src/main/java/org/openobservatory/ooniprobe/model/NetworkMeasurement.java
lorenzoPrimi: I don't need the long description. Only test name is stored in old versions

app/src/main/AndroidManifest.xml
hellais: Yes I think we also want to support `http` links. I don't think it's a security concern as actually if the app is installed it will never hit the network. We do generate them by default as `https`, but I think it's best in the app to support both to avoid the case where users mistakenly use a http link and we will only trigger the app upon redirect (so if `ooni.io` is blocked nothing will be triggered)

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
hellais: Can we add a check to validate the URLs (by parsing the JSON) when the test name parameter is validated? This way we could show an error as early on as possible.

hellais: I think that if we are at this point of the code and the JSON is malformed, we should not proceed with running the test, but rather should show an error message of some sort and abort running the test.

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
hellais: @lorenzoPrimi can you then at least put a comment with a link to where you got the code snippet from?

app/src/main/java/org/openobservatory/ooniprobe/data/TestData.java
hellais: Totally agree. We are also using the magic number 2 in other places and we could define these constants in one unique place and put them there.

app/src/main/java/org/openobservatory/ooniprobe/model/NetworkMeasurement.java
hellais: https://en.wikipedia.org/wiki/Cargo_cult_programming

app/src/main/java/org/openobservatory/ooniprobe/utils/AlarmReceiver.java
hellais: We want to run at minimum "http_header_field_manipulation" too. Maybe also "dash". I would say that as it is now, we should probably have a list of all the supported tests somewhere (it probably already exists) and just run them all.

ChangeLog.md
hellais: Since 1.2.0 was already cut from what I can see. The real release is actually going to be 1.2.1 and this should be listed in the changelog as a YANKED version.

app/src/main/AndroidManifest.xml
hellais: Does XML not support comments in the form `<!-- This is a comment -->`. Why is it considered a bad practice? It's fairly common in HTML to do it.

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
lorenzoPrimi: Here are the definitions:
https://developer.android.com/reference/android/content/Intent.html

I guess this one is the snippet
https://stackoverflow.com/questions/14332441/finish-current-activity-and-start-a-new-one

hellais: @lorenzoPrimi excellent! Can you put that in a comment in the code?

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
lorenzoPrimi: What should happen if the JSON is malformed? I can show a popup, and then?

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
bassosimone: > Either MainActivity or RunTestActivity will be running

Okay. And what happens when you quit `RunTestActivity`?

BTW: long time ago I remember discussing with @antoniolangiu that Android apps should use intents to communicate between backend services and user-facing activities. That was something that @antoniolangiu knew and it was not clear to me where this knowledge come from. Maybe, @antoniolangiu, can you share with us a couple of URLs related to that, so we can read a little bit, and possibly planning a refactoring oriented to using intents?

app/src/main/java/org/openobservatory/ooniprobe/model/NetworkMeasurement.java
bassosimone: LOL, I did not think at that! My bad! :-D

app/src/main/java/org/openobservatory/ooniprobe/model/NetworkMeasurement.java
bassosimone: Thanks!

ChangeLog.md
```diff
@@ -1,4 +1,4 @@
-# ooniprobe-android 1.2.0 [2017-09-25]
+# ooniprobe-android 1.2.1 [2017-09-25]
```
hellais: Shouldn't this be 1.2.0?

app/src/main/java/org/openobservatory/ooniprobe/activity/MainActivity.java
```diff
@@ -71,6 +72,8 @@
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
 
+System.out.println("VERSION: "+VersionUtils.get_software_version());
```
hellais: I think this is a debug line that you left in here.

app/src/main/java/org/openobservatory/ooniprobe/utils/VersionUtils.java
```diff
@@ -0,0 +1,11 @@
+package org.openobservatory.ooniprobe.utils;
+
+import org.openobservatory.ooniprobe.BuildConfig;
+
+public class VersionUtils {
+
+    public static String get_software_version(){
+        String release_name = "-rc.5";
```
hellais: Can we put this in a place where we have defines?

ChangeLog.md
lorenzoPrimi: My bad I shouldn't have pushed the Changelog already, it contains the draft for the 1.2.1. updating

app/src/main/java/org/openobservatory/ooniprobe/utils/VersionUtils.java
lorenzoPrimi: We have the defines in model.OONITests.

hellais: Cool, then put the global version information in there.

hellais: Or maybe put it in the manifest like this: https://stackoverflow.com/a/7929059

lorenzoPrimi: Even better, I moved it in the build.gradle file

app/src/main/java/org/openobservatory/ooniprobe/activity/RunTestActivity.java
```diff
@@ -0,0 +1,265 @@
...
+                JSONArray urls = taJson.getJSONArray("urls");
+                for (int i = 0; i < urls.length(); i++)
+                    urlItems.add(urls.getString(i));
+            } catch (JSONException e) {
+                Alert.alertDialogWithAction(this, getString(R.string.invalid_parameter), getString(R.string.urls), new DialogInterface.OnClickListener() {
+                    public void onClick(DialogInterface dialog, int id) {
+                        goToMainActivity();
+                    }
+                });
+                e.printStackTrace();
```
bassosimone: I don't think it's correct to continue after the exception. You should probably return here, no?

lorenzoPrimi: I show an alert and close the activity as you requested

bassosimone: Well, but then the code falls through and you call:

- `listItems.addAll(urlItems);`
- then there is a series of ifs that modify `mUrlListAdapter`
- then you call `runButton.OnClickListener`
- then ...

My point is: if, on error, you want to _stop_ the execution of the current function, you should `return`.

I am missing something obvious?

lorenzoPrimi: Now I understand you, I call finish() only when the user clicks OK in the alert

bassosimone:
@lorenzoPrimi please go ahead with merging. 🐳

PR #117, opened 2017-09-28
Solving #112

### LINKED ISSUE #112: The run view persists if not dismissed
The bug reported in: https://github.com/TheTorProject/ooniprobe-ios/issues/132 also happens on Android in `1.2.2-rc.1`

lorenzoPrimi:
Needs more testing, but I guess it's ok

Closes #112

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -14,8 +14,8 @@ android {
         applicationId "org.openobservatory.ooniprobe"
         minSdkVersion 14
         targetSdkVersion 25
-        versionCode 16
-        versionName "1.2.2"
+        versionCode 17
+        versionName "1.2.3-alpha.1"
         setProperty("archivesBaseName", "ooniprobe-$versionName")
     }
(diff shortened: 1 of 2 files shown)


bassosimone:
The diff is very simple, but to properly sign-off this PR I also need to test on the emulator or on a device and I will not be able to do that until this afternoon.

bassosimone:
I am now testing this

bassosimone:
Okay, so, I have tested this and it works 👍. @lorenzoPrimi can you explain me why? (By just reading the diff I am not able to fully understand)

bassosimone:
> By just reading the diff I am not able to fully understand)

Is it because the same activity - already open - can be activated by a new intent? So, having an handler for a new intent allows to deal with the screen drawing another test?

lorenzoPrimi:
onNewIntent() is triggered only when the activity is already opened and you send it a new intent (so new parameters.

PR #118, opened 2017-09-28
Fix/results list

lorenzoPrimi:


app/src/main/java/org/openobservatory/ooniprobe/fragment/ResultFragment.java
```diff
@@ -68,6 +68,7 @@ public View onCreateView(LayoutInflater inflater, ViewGroup container,
             wv.loadUrl("file:///android_asset/webui/index.html");
         } catch (IOException e) {
             // XXX log the IO exception in some way
+            // It should never happens because we already opened the file when arriving at this point
```
hellais: Then maybe the `File` call needs to also be wrapped in the `try` catch block. I don't see this handled in other way :|

lorenzoPrimi: Why the IDE doesn't force me? maybe just declaring it doesn't throw an Exception? @bassosimone

bassosimone:
I will look into this later this afternoon.

app/src/main/java/org/openobservatory/ooniprobe/fragment/ResultFragment.java
bassosimone: @lorenzoPrimi good question. So, my understanding [of the documentation](https://developer.android.com/reference/java/io/File.html#File(java.lang.String,%20java.lang.String)) is that the constructor only throws when the input path is `nullptr`, therefore I assume the constructor is lazy and I/O is attempted later.

PR #124, opened 2017-10-16
Edited the mobile app description

agrabeli:
@lorenzoPrimi @hellais I've updated/edited the app description (the same has been done for f-droid).

### DIFF
--- a/description/source.xlf
+++ b/description/source.xlf
@@ -13,7 +13,7 @@
         <note from="developer">This is limited by Google to 80 characters</note>
       </trans-unit>
       <trans-unit id="title-ios" maxwidth="50" size-unit="char">
-          <source>ooniprobe - Measure Internet Censorship &amp; Speed</source>
+          <source>OONI Probe - Measure Internet Censorship &amp; Speed</source>
           <note from="developer">This is limited by Apple to 50 characters</note>
       </trans-unit>
       <trans-unit id="subtitle-ios" maxwidth="30" size-unit="char">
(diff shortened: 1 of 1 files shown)


lorenzoPrimi:
Should I modify the same file in the -iOS repo?

hellais:
> Should I modify the same file in the -iOS repo?

I think we should probably have a single source of truth for these descriptions. I would suggest we consolidate all mobile app descriptions in one or the other repo and remove the duplication.

lorenzoPrimi:
Yeah, and remove the description files from both repos

hellais:
> Yeah, and remove the description files from both repos

humm, you probably want to keep it somewhere. Unless you have an idea of some other repo where the description should live.

lorenzoPrimi:
The options we have are two:
1) Create a repo with descriptions and script to generate them
2) Keep them in bot repo (iOS - Android)

I guess 1 is the cleanest

lorenzoPrimi:
@hellais  How do I name this new repo? where do I create it?
Or you want to create it and I'll commit files?

hellais:
> @hellais How do I name this new repo? where do I create it?
> Or you want to create it and I'll commit files?

Let's just stick with keeping it in here (the android repo) and remove it from the iOS repo. This tool could maybe also be useful to make this task easier: https://docs.fastlane.tools/actions/supply/

lorenzoPrimi:
Ok I will remove the files from the iOS repo and keep them here.
Can you modify the file scripts/gen-descriptions.py to accept ios or android param and generate the correspondent strings? @hellais

hellais:
> Can you modify the file scripts/gen-descriptions.py to accept ios or android param and generate the correspondent strings?

Maybe the easy fix for this is just to create two `gen-descriptions.py` scripts called `gen-descriptions-ios.py` and `gen-descriptions-android.py`. Also did you look at fastlane? It seems like it would make this task much easier and automated.

hellais:
@lorenzoPrimi I filed this ticket about that: https://github.com/TheTorProject/ooniprobe-android/issues/125

lorenzoPrimi:
Yes seen it, but I don't find it useful for me now.

hellais:
> Yes seen it, but I don't find it useful for me now.

Well it avoids you having to copy paste (and possibly making mistakes in doing so) all the descriptions and metadata for every possible language.

Is this not useful?

lorenzoPrimi:
I said "for now" like there are other priorities, and the description is updated not often. 
In fact I will try it after the revamp that will need new screenshots and description.

hellais:
> I said "for now" like there are other priorities, and the description is updated not often.

This PR you just merged actually updates the description, so you are going to have to update the description for all the languages once transifex translation converges.

lorenzoPrimi:
Remember that on iOS to update the description we should upload a new build and there is no schedule release for now.

PR #135, opened 2017-12-14
fixing issue#108

TeMuNiDe:
As this app uses an application class, Firebase may not initialize itself, So it need to be initialized in the application class.


lorenzoPrimi:
Thanks for pointing it out

PR #137, opened 2018-01-05
Feature/im tests

lorenzoPrimi:


app/src/main/java/org/openobservatory/ooniprobe/data/TestData.java
```diff
@@ -313,6 +347,113 @@ public static void setAnomaly_ndt(String entry, NetworkMeasurement test){
...
+                if (telegram_web_status == null)
+                    anomaly = OONITests.ANOMALY_ORANGE;
+                else if (telegram_web_status.equals("blocked"))
+                    anomaly = OONITests.ANOMALY_RED;
+            }
+            if (test.anomaly < anomaly) {
+                test.anomaly = anomaly;
+                TestStorage.setAnomaly(context, test.test_id, anomaly);
+            }
+        } catch (JSONException e) {
```
bassosimone: Maybe here you want to log the exception, at least?

app/src/main/java/org/openobservatory/ooniprobe/data/TestData.java
```diff
@@ -313,6 +347,113 @@ public static void setAnomaly_ndt(String entry, NetworkMeasurement test){
...
+        }
+
+
+    }
+
+    /*
+    FB: red blocking if either "facebook_tcp_blocking" or "facebook_dns_blocking" is true
+    docs: https://github.com/TheTorProject/ooni-spec/blob/master/test-specs/ts-019-facebook-messenger.md#semantics
+    */
+    public static void setAnomaly_facebook_messenger(String entry, NetworkMeasurement test) {
```
bassosimone: The three new functions are all similar. I suggest considering writing a single function with more parameters, as it seems the main difference is the list of keys to check for.

lorenzoPrimi: These have been tested and working, I suggest to don't touch them for this release as in the next everything is gonna be changed

app/src/main/java/org/openobservatory/ooniprobe/data/TestData.java
lorenzoPrimi: done

app/src/main/java/org/openobservatory/ooniprobe/data/TestData.java
```diff
@@ -313,6 +347,113 @@ public static void setAnomaly_ndt(String entry, NetworkMeasurement test){
...
+                    anomaly = OONITests.ANOMALY_ORANGE;
+                else if (value.equals("blocked"))
+                    anomaly = OONITests.ANOMALY_RED;
+            }
+            if (test.anomaly < anomaly) {
+                test.anomaly = anomaly;
+                TestStorage.setAnomaly(context, test.test_id, anomaly);
+            }
+        } catch (JSONException e) {
+            System.out.println("JSONException "+ e);
```
hellais: Why are you using `System.out.println` here and `Log.v` above?

I would suggest using always the same functions for logging as it makes debugging logs easier.

app/src/main/java/org/openobservatory/ooniprobe/data/TestData.java
```diff
@@ -313,6 +347,114 @@ public static void setAnomaly_ndt(String entry, NetworkMeasurement test){
...
+                    anomaly = OONITests.ANOMALY_ORANGE;
+                else if (telegram_web_status.equals("blocked"))
+                    anomaly = OONITests.ANOMALY_RED;
+            }
+            if (test.anomaly < anomaly) {
+                test.anomaly = anomaly;
+                TestStorage.setAnomaly(context, test.test_id, anomaly);
+            }
+        } catch (JSONException e) {
+            System.out.println("JSONException "+ e);
```
hellais: Same comment as above.

app/src/main/java/org/openobservatory/ooniprobe/data/TestData.java
```diff
@@ -313,6 +347,114 @@ public static void setAnomaly_ndt(String entry, NetworkMeasurement test){
         }
     }
 
+    /*
+     whatsapp: red if "whatsapp_endpoints_status" or "whatsapp_web_status" or "registration_server" are "blocked"
+     docs: https://github.com/TheTorProject/ooni-spec/blob/master/test-specs/ts-018-whatsapp.md#semantics
+     */
+    public static void setAnomaly_whatsapp(String entry, NetworkMeasurement test) {
```
hellais: My eyes! `mixedCamel_and_snake_Case`

app/src/main/java/org/openobservatory/ooniprobe/data/TestData.java
lorenzoPrimi: basically when it's Log.v it was Antonio's code, system.out is mine. I replaced them all.

app/src/main/java/org/openobservatory/ooniprobe/data/TestData.java
lorenzoPrimi: ehehe because is was using functionName when it was only setAnomaly, then i added test_name 😅

lorenzoPrimi: As every function is mixedCamel, what do you suggest to use for them? to use testName even if everywhere the vars are called test_name? @hellais

app/src/main/java/org/openobservatory/ooniprobe/data/TestData.java
bassosimone: Is there a reason to prefer `System.out` to `Log.v`?

app/src/main/java/org/openobservatory/ooniprobe/data/TestData.java
bassosimone: I _think_ we might want to use `camelCase` for methods, `PascalCase` for classes. IIRC on Java usually people uses `camelCase` also for attributes. I would say we can do more or less what we want with private attributes, as long as we strive for consistency.

app/src/main/java/org/openobservatory/ooniprobe/data/TestData.java
bassosimone: I must admin that I am not comfortable with:

1. blocking someone else work because of suggested refactorings (hence, I'd say let's go forward with this code and okay for improve stuff in next release)

2. the concept that _everything_ is changed by next release (I would be happier to see more incremental improvement where possible; that said, I also understand that apps are very tightly coupled to the UI... but still rewriting the world makes me nervous)

app/src/main/java/org/openobservatory/ooniprobe/data/TestData.java
lorenzoPrimi: I even believe the setAnomaly function should be only one with test name as param

hellais: I think the correct name for a function like this one should be `setAnomalyWhatsapp`

lorenzoPrimi: and for `setAnomaly_facebook_messenger` ? should I use `setAnomalyFacebookMessenger` even if the test is called `facebook_messenger`?

hellais: Yes correct `setAnomalyFacebookMessenger`, well the test is technically called "Facebook Messenger", `facebook_messenger` is the snake case version of it, while `facebookMessenger` is the camelCase capitalisation of it.

PR #159, opened 2018-11-08
Update/mk

lorenzoPrimi:


app/src/main/java/org/openobservatory/ooniprobe/model/settings/Settings.java
```diff
@@ -62,18 +64,19 @@ public Settings(Context c, PreferenceManager pm) {
...
+			geoip_asn_path = c.getCacheDir() + "/" + Application.ASN_MMDB;
+			geoip_country_path = c.getCacheDir() + "/" + Application.COUNTRY_MMDB;
+			no_collector = pm.isUploadResults();
+			save_real_probe_asn = pm.isIncludeAsn();
+			save_real_probe_cc = pm.isIncludeCc();
+			save_real_probe_ip = pm.isIncludeIp();
 			software_name = "ooniprobe-android";
 			software_version = VersionUtils.get_software_version();
-			randomize_input = 0;
+			randomize_input = false;
```
bassosimone: So, no-one is using `randomize_input`. I'd like to get rid of it in MK. It sounds like extra complexity now.

app/src/main/java/org/openobservatory/ooniprobe/utils/NotificationService.java
```diff
@@ -78,56 +78,27 @@ public void sendRegistrationToServer() {
...
+		//client.setAvailableBandwidth(String value);
+		client.setDeviceToken(device_token);
+		client.setCABundlePath(ca_bundle_path);
+		client.setGeoIPCountryPath(geoip_country_path);
+		client.setGeoIPASNPath(geoip_asn_path);
+		client.setLanguage(language);
+		client.setNetworkType(network_type);
+		client.setPlatform("android");
+		//TODO what to add here?
+		//client.setProbeFamily(String value);
```
bassosimone: We do not use this field presently.

app/src/main/java/org/openobservatory/ooniprobe/utils/NotificationService.java
```diff
@@ -78,56 +78,27 @@ public void sendRegistrationToServer() {
...
+		client.setProbeTimezone(TimeZone.getDefault().getDisplayName(true, TimeZone.SHORT));
```
bassosimone: What string would that be?

app/src/main/java/org/openobservatory/ooniprobe/utils/NotificationService.java
```diff
@@ -78,56 +78,27 @@ public void sendRegistrationToServer() {
...
+		client.setRegistryURL(NOTIFICATION_SERVER);
+		client.setSecretsFile(auth_secret_file);
+		client.setSoftwareName(software_name);
+		client.setSoftwareVersion(software_version);
+		//TODO add @"web_connectivity", @"whatsapp", @"telegram", @"facebook_messenger", @"ndt", @"dash", @"http_invalid_request_line", @"http_header_field_manipulation"
+		//client.addSupportedTest();
```
bassosimone: Yeah, I suggest doing that :-)

app/src/main/java/org/openobservatory/ooniprobe/utils/NotificationService.java
lorenzoPrimi: Should be "UTC" or "CEST" or the short version that doesn't include the Country/City

bassosimone: Unless this was discussed and I missed the discussion, I would suggest to include the Country/City because it is information useful to us. Isn't it?

PR #163, opened 2018-11-19
Delete scripts and files that have been migrated to ooni/translations

hellais:
All app translations shall from now on be managed via: https://github.com/ooni/translations.

In this branch I delete all the deprecated code to avoid duplication.


hellais:
I have also update the `update_languages.sh` script to use the new and improved workflow.

lorenzoPrimi:
This PR and the relative branch be deleted.
Implemented in https://github.com/ooni/translations/pull/6

PR #189, opened 2019-02-01
Fix bug in uploading of measurements

### LINKED ISSUE #188: Report ID field is empty in the raw data section
## Expected Behavior

When I go into the view data section of the app I should see in the `report_id` populated with a value.

This does not happen on probe-ios.

## Actual Behavior

The `report_id` field contains the empty strings `""`.

See:
signal-attachment-2019-01-31-104928

## Steps to Reproduce the Problem

  1. Run any experiment on android
  2. Go the view data section of the measurement via the option menu
  3. Notice how the report_id field is empty

## Specifications

  - Version: 2.0.0
  - Platform: android

hellais:
closes #188


hellais:
Superseeded by #190

PR #203, opened 2019-03-08
closing ooni/probe-ios#234 in the proper way

lorenzoPrimi:
Fixes ooni/probe-ios#234


app/src/main/java/org/openobservatory/ooniprobe/test/test/AbstractTest.java
```diff
@@ -54,6 +55,7 @@ void run(Context c, PreferenceManager pm, Gson gson, Settings settings, Result r
 		settings.name = mkName;
 		settings.inputs = inputs;
 		settings.options.max_runtime = max_runtime;
+		settings.annotations.origin = origin;
```
lorenzoPrimi: what happens when origin is null? is this annotation being put in the json?

lorenzoPrimi: seems it is not put there, ok

lorenzoPrimi:
Not merging because the commit is in another branch

PR #214, opened 2019-04-06
Feature/upload result manually

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/838

Files important to review:
- MKCollectorResubmitSettings.java
- PreferenceManager.java
- MeasurementItem.java
- Measurement.java

PR #217, opened 2019-04-18
add MKCollectorResubmitSettings async task

xanscale:
Is business logic for https://github.com/ooni/probe/issues/838

Important file:
MKCollectorResubmitTask


app/src/main/res/values/untraslatable.xml
```diff
@@ -29,6 +29,7 @@
 	<string name="bold" translatable="false">&lt;b>%1$s&lt;/b></string>
 	<string name="normalBold" translatable="false">%1$s &lt;b>%2$s&lt;/b></string>
 	<string name="bigNormal" translatable="false">%1$s&lt;small>&lt;small>&lt;small>%2$s&lt;/small>&lt;/small>&lt;/small></string>
+	<string name="paramOfParam" translatable="false">%1$s / %2$s</string>
```
bassosimone: What's that?

bassosimone: Ah, that's the "uncommitted res". Oops!

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
```diff
@@ -0,0 +1,87 @@
...
+	 *
+	 * @param activity from this task are executed
+	 */
+	public MKCollectorResubmitTask(A activity) {
+		super(activity, true, false);
+	}
+
+	private static void perform(Context c, Measurement m) throws IOException {
+		FileInputStream is = new FileInputStream(Measurement.getEntryFile(c, m.id, m.test_name));
+		String input = new GsonBuilder().disableHtmlEscaping().create().toJson(new JsonParser().parse(new InputStreamReader(is)));
```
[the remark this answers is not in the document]
xanscale: validazione e conversione a stringa

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
```diff
@@ -0,0 +1,87 @@
...
+			throw new IllegalArgumentException("MKCollectorResubmitTask require 2 nullable params: result_id, measurement_id");
+		List<Measurement> measurements = MeasurementDao.queryList(params[0], params[1], false, false);
+		for (int i = 0; i < measurements.size() && getActivity() != null; i++) {
+			publishProgress(getActivity().getString(R.string.Modal_ResultsNotUploaded_Uploading, getActivity().getString(R.string.paramOfParam, Integer.toString(i + 1), Integer.toString(measurements.size()))));
+			Measurement m = measurements.get(i);
+			m.result.load();
+			try {
+				perform(getActivity(), m);
+			} catch (IOException e) {
+				e.printStackTrace();
```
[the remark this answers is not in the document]
xanscale: yes

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
```diff
@@ -0,0 +1,87 @@
...
+			m.save();
+		} else {
+			Log.w(MKCollectorResubmitSettings.class.getSimpleName(), results.getLogs());
+			// TODO check what to do with logs
+		}
+	}
+
+	@Override protected void onPreExecute() {
+		super.onPreExecute();
+		if (getActivity() != null)
```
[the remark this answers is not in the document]
xanscale: lifecycle os management start only on idle of the main thread. so in this case there is no difference.

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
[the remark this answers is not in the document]
bassosimone: The code that follows cannot avoid parsing the measurement because it needs to update the report ID.

For this reason I think we can skip this step here. If you don’t want to skip this step, then you should probably also validate the returned measurement to make sure that it is also JSON.

I do not have a strong opinion on what it is better to do. Maybe doing less work implies draining less battery. At the same time, more work implies more robustness (we used to have a bug where MK did not always emit valid JSON and this strategy would have helped to catch that).

I leave to you go choose the strategy you think it’s better here, just let’s be consistent.

Actually, another thing to keep in mind is whether we want to validate a measurement when it is emitted by MK rather than here. OTOH a measurement may be corrupted on disk, so from a conceptual and robustness oriented point of view, one should always validate what MK gives us.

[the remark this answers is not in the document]
bassosimone: The code that follows cannot avoid parsing the measurement because it needs to update the report ID.

For this reason I think we can skip this step here. If you don’t want to skip this step, then you should probably also validate the returned measurement to make sure that it is also JSON.

I do not have a strong opinion on what it is better to do. Maybe doing less work implies draining less battery. At the same time, more work implies more robustness (we used to have a bug where MK did not always emit valid JSON and this strategy would have helped to catch that).

I leave to you go choose the strategy you think it’s better here, just let’s be consistent.

Actually, another thing to keep in mind is whether we want to validate a measurement when it is emitted by MK rather than here. OTOH a measurement may be corrupted on disk, so from a conceptual and robustness oriented point of view, one should always validate what MK gives us.

app/src/main/res/values/untraslatable.xml
xanscale: is only for create a "3 / 42" string

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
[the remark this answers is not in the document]
bassosimone: I’d rather write robust code that requires less knowledge anyways, especially when doing that is easy and especially when the result requires the reader to know less, not more, about the internals of the systems and about whether they did change.

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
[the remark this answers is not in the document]
bassosimone: Ok then we’re all good here

app/src/main/java/org/openobservatory/ooniprobe/dao/MeasurementDao.java
```diff
@@ -0,0 +1,36 @@
...
+import org.openobservatory.ooniprobe.model.database.Measurement;
+import org.openobservatory.ooniprobe.model.database.Measurement_Table;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import androidx.annotation.Nullable;
+
+public class MeasurementDao {
+	public static List<Measurement> queryList(@Nullable Integer resultId, @Nullable Integer measurementId, @Nullable Boolean uploaded, @Nullable Boolean failed) {
```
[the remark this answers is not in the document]
xanscale: i thinked this method to use on all cases we need a measurement list from DB, using params to add where clausole if they are not null.

anyway first 2 params will called as null in some cases now.

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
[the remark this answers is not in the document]
xanscale: actually i just checked file corruption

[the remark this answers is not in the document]
bassosimone: Right, so I guess it also makes sense to verify what MK passes us, no?

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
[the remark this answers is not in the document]
bassosimone: Please, use this pattern. I believe in general it's more OOP, because we try and then see if we could rather than requesting permission beforehand! :-)

app/src/main/java/org/openobservatory/ooniprobe/dao/MeasurementDao.java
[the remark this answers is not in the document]
bassosimone: > anyway first 2 params will called as null in some cases now.

 In this PR? Did I miss it?

gabelula:
@xanscale Can you add a description to this PR and what you think needs to be reviewed (example at https://github.com/ooni/probe-android/pull/214) ?

xanscale:
Done

app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
```diff
@@ -130,20 +129,17 @@ public static Intent newIntent(Context context, int id) {
 		switch (item.getItemId()) {
 			case R.id.rawData:
 				try {
-					FileInputStream is = new FileInputStream(Measurement.getEntryFile(this, measurement.id, measurement.test_name));
-					String json = new GsonBuilder().setPrettyPrinting().disableHtmlEscaping().create().toJson(new JsonParser().parse(new InputStreamReader(is)).getAsJsonObject());
+					String json = FileUtils.readFileToString(Measurement.getEntryFile(this, measurement.id, measurement.test_name), StandardCharsets.UTF_8);
```
hellais: In looking at the code behind `getEntryFile` I found it quite surprising that:
```
	static File getMeasurementDir(Context c) {
		File dir = new File(c.getFilesDir(), Measurement.class.getSimpleName());
		dir.mkdirs();
		return dir;
	}
```

is calling `dir.mkdirs` which is a side-effect I would not expect from a getter method.

Moreover according to the android docs: https://developer.android.com/reference/android/content/Context.html#getFilesDir()

> The returned path may change over time if the calling app is moved to an adopted storage device, so only relative paths should be persisted.

Hence repeated calls to `getMeasurementDir` may not yield the same value and create unexpected side effects if not properly controlled.

My suggestion would be to at least explicitly call the directory tree creation function separately from determining the measurement filepath.

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
```diff
@@ -0,0 +1,84 @@
...
+			activity.getWindow().addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
+	}
+
+	@Override protected Void doInBackground(Integer... params) {
+		if (params.length != 2)
+			throw new IllegalArgumentException("MKCollectorResubmitTask requires 2 nullable params: result_id, measurement_id");
+		List<Measurement> measurements = MeasurementDao.queryList(params[0], params[1], false, false);
+		for (int i = 0; i < measurements.size(); i++) {
+			A activity = getActivity();
+			if (activity != null) {
```
hellais: This can be refactored as:
```
if (activity == null) {
  continue
}
[ the rest ]
```

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
```diff
@@ -0,0 +1,84 @@
...
+				publishProgress(activity.getString(R.string.Modal_ResultsNotUploaded_Uploading, activity.getString(R.string.paramOfParam, Integer.toString(i + 1), Integer.toString(measurements.size()))));
```
hellais: If I understand this correctly you are basically formatting the string that is going to show the progress line.

Perhaps you can split the line used to generate the progress string from the actual posting of it (is there perhaps a format string function in java?) into two lines and comment above it how it looks like.

app/src/main/java/org/openobservatory/ooniprobe/dao/MeasurementDao.java
```diff
@@ -0,0 +1,36 @@
...
+
+import org.openobservatory.ooniprobe.model.database.Measurement;
+import org.openobservatory.ooniprobe.model.database.Measurement_Table;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import androidx.annotation.Nullable;
+
+public class MeasurementDao {
```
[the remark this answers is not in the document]
hellais: This is still missing.

app/src/main/java/org/openobservatory/ooniprobe/dao/MeasurementDao.java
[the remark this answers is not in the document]
hellais: I agree with @bassosimone I don't see the need to abstract on top of something which already is an abstraction on top of SQL.

Effectively you are using the arguments to this function call to determine how the SQL filtering should happen, yet using directly the ORM semantics is much more expressive. I think that since you have chosen to use a query builder already you should be using that.

Moreover you are calling this method in the same way (`queryList()`) as the query builder method which I think creates some confusion.

I would suggest dropping this method and moving this code directly into the single consume of it (and perhaps simplifying it a bit which it can given that it's only a single use).

app/src/main/java/org/openobservatory/ooniprobe/test/test/AbstractTest.java
```diff
@@ -87,10 +86,7 @@ void run(Context c, PreferenceManager pm, Gson gson, Settings settings, Result r
 						measurement.save();
 						break;
 					case "log":
-						if (logFOS == null)
-							logFOS = new FileOutputStream(Measurement.getLogFile(c, result.id, name));
-						logFOS.write(event.value.message.getBytes());
-						logFOS.write('\n');
+						FileUtils.writeStringToFile(Measurement.getLogFile(c, result.id, name), event.value.message + "\n", StandardCharsets.UTF_8, true);
```
hellais: Maybe adding a comment saying that `true` means we are appending to the file is useful.

Though this may just be my lack of familiarity with these java function calls.

app/src/main/java/org/openobservatory/ooniprobe/dao/MeasurementDao.java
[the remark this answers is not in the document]
xanscale: removed DAO

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
```diff
@@ -73,16 +73,16 @@ private static void perform(Context c, Measurement m) throws IOException {
...
-				publishProgress(activity.getString(R.string.Modal_ResultsNotUploaded_Uploading, paramOfParam));
-				Measurement m = measurements.get(i);
-				m.result.load();
-				try {
-					perform(activity, m);
-				} catch (IOException e) {
-					e.printStackTrace();
-				}
+			if (activity == null)
+				break;
```
hellais: This is not equivalent to how it was before. To be equivalent you should do `continue` when it's null.

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
```diff
@@ -73,16 +73,16 @@ private static void perform(Context c, Measurement m) throws IOException {
...
+			String paramOfParam = activity.getString(R.string.paramOfParam, Integer.toString(i + 1), Integer.toString(measurements.size()));
+			publishProgress(activity.getString(R.string.Modal_ResultsNotUploaded_Uploading, paramOfParam));
+			Measurement m = measurements.get(i);
+			m.result.load();
+			try {
+				perform(activity, m);
+			} catch (IOException e) {
+				e.printStackTrace();
```
hellais: Please use more descriptive git commit messages.

`little clean` is not a descriptive commit message.

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
xanscale: i know, in that case break is better

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
xanscale: ok

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
hellais: Why? In which cases can the activity be null?

I now have more question then:
* Why are we creating multiple instances of the activity on every cycle loop and not just creating a single one at the beginning of the loop?
* In which cases will the activity be null on an iteration of the loop? Should we perhaps not throw an exception in that case?

app/src/main/java/org/openobservatory/ooniprobe/model/database/Result.java
```diff
@@ -58,7 +58,7 @@ public static Result getLastResult(String test_group_name) {
 
 	public static void deleteAll(Context c) {
 		try {
-			FileUtils.cleanDirectory(Measurement.getMeasurementDir(c));
+			FileUtils.cleanDirectory(Measurement.createMeasurementDir(c));
```
hellais: I didn't actually realise in all the places in which this is being called. I think that since it's being used also in situations where it doesn't make sense for it to be called this way, then I suggest you do as I originally suggested and split the creation of the directory from getting the pathname.

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
```diff
@@ -0,0 +1,98 @@
...
+		if (params.length != 2)
+			throw new IllegalArgumentException("MKCollectorResubmitTask requires 2 nullable params: result_id, measurement_id");
+		ArrayList<SQLOperator> where = new ArrayList<>();
+		if (params[0] != null)
+			where.add(Measurement_Table.result_id.eq(params[0]));
+		if (params[1] != null)
+			where.add(Measurement_Table.id.eq(params[1]));
+		where.add(OperatorGroup.clause().or(Measurement_Table.is_uploaded.eq(false)).or(Measurement_Table.report_id.isNull()));
+		where.add(Measurement_Table.is_failed.eq(false));
+		List<Measurement> measurements = SQLite.select().from(Measurement.class).where(where.toArray(new SQLOperator[0])).queryList();
```
hellais: Can you reindent this line to make it a bit more readable?

hellais: I also don't fully understand why you are doing `new SQLOperator[0]`. What are these indexes?

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
```diff
@@ -0,0 +1,98 @@
...
+			activity.getWindow().addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
+	}
+
+	@Override protected Void doInBackground(Integer... params) {
+		if (params.length != 2)
+			throw new IllegalArgumentException("MKCollectorResubmitTask requires 2 nullable params: result_id, measurement_id");
+		ArrayList<SQLOperator> where = new ArrayList<>();
+		if (params[0] != null)
+			where.add(Measurement_Table.result_id.eq(params[0]));
+		if (params[1] != null)
```
hellais: Will this ever be called with a `params` array which has two null entries? If not we should maybe explicitly check this at the beginning and throw an expection if that is the case.

I think this would also allow us to get rid of this `ArrayList` for the where clause and reduce complexity.

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
xanscale: we do not create multiple instances, only geting multiple times same instance.
we do that because use can close app during process, and in this way app will closed and not killed

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
```diff
@@ -0,0 +1,98 @@
...
+			where.add(Measurement_Table.id.eq(params[1]));
+		where.add(OperatorGroup.clause().or(Measurement_Table.is_uploaded.eq(false)).or(Measurement_Table.report_id.isNull()));
```
hellais: We should document why `is_uploaded` being false is equivalent to `report_id` being null in a comment.

app/src/main/java/org/openobservatory/ooniprobe/model/database/Result.java
xanscale: we need to talk about that

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
xanscale: it will be called in all combination

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
xanscale: due to MK lib bug

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
hellais: > we do not create multiple instances, only geting multiple times same instance.

Ah yes that's a good point.

> we do that because use can close app during process, and in this way app will closed and not killed

 Can you elaborate on this and better explain what you mean by this?

In which cases will the activity be null?

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
hellais: > it will be called in all combination

You mean that it's possible for `param[0]` to be `null`? In which case will that happen?

Maybe what you can do if that is the case is define a `result_id` variable and then use the syntax sugar of your ORM to form the query directly without passing via an array.

Example:

```
		if (params.length != 2)
			throw new IllegalArgumentException("MKCollectorResubmitTask requires 2 nullable params: result_id, measurement_id");
		String result_id;
		if (params[0] != null) {
                    result_id = params[0]
                 } else if (params[1] != null) {
                     result_id = params[1]
                } else {
                    // XXX is this even possible? should we maybe throw an exception?
                }
		where.add(.or().or());
		where.add();
		List<Measurement> measurements = SQLite.select().from(Measurement.class)
                 // We check on both the report_id and is_uploaded as we may have some unuploaded measurements which are marked as is_uploaded = true, but we always know that those with report_id set to null are not uploaded
                  .and(
                      OperatorGroup.clause().or(Measurement_Table.is_uploaded.eq(false))
                      .or(Measurement_Table.report_id.isNull()
                  )
		.and(Measurement_Table.result_id.eq(result_id))
                 .queryList();

```

app/src/main/java/org/openobservatory/ooniprobe/model/database/Result.java
xanscale: in this way i don't care if the folder exists or not. i can use abstracting from that. what do you mean in particular ?

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
xanscale: indent is correct

hellais: > indent is correct

Does you github look different from mine?

Screenshot 2019-04-23 at 19 56 08

app/src/main/java/org/openobservatory/ooniprobe/model/database/Result.java
hellais: I mean that if you read that code above it doesn't make any sense that you are `createMeasurementDir` then to delete it. So the function name should be reflected in that no?

xanscale: So maybecreate or back to get

hellais: As I have already written in a couple of places

> I suggest you do as I originally suggested and split the creation of the directory from getting the pathname

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
xanscale: When user close app during upload

hellais: And by close you mean the user puts the app in background or they actually force close it?

What happens to measurements that are being uploaded? Is there a possibility of a race condition or of the DB ending up in an inconsistent state?

What happens if the user re-opens the app, are they able to resume the upload?

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
xanscale: params[0] are result_id
params[1] are measurement_id

Null null means get all measurements

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
xanscale: Force close,

I made each measurement upload atomic, so on second run start on remaining measurements

app/src/main/java/org/openobservatory/ooniprobe/model/database/Result.java
xanscale: I use this compact way to abstract from know if folder exists or not in every place I use it. How can I do same?

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
xanscale: Sorry I not understand correctly. I will fix

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	@Override protected void onPreExecute() {
+		super.onPreExecute();
+		A activity = getActivity();
+		if (activity != null)
+			activity.getWindow().addFlags(WindowManager.LayoutParams.FLAG_KEEP_SCREEN_ON);
+	}
+
+	@Override protected Void doInBackground(Integer... params) {
```
hellais: A docstring should be added explaining that this function takes two positional arguments which are one the `measurement_id` and the other is the `result_id`. If it has both arguments set to `null` it will then upload all measurements.

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
hellais: Ok I understand now. Thanks for explaining.

How about then doing something like:
```
	 msmQuery = SQLite.select().from(Measurement.class)
                 // We check on both the report_id and is_uploaded as we 
                // may have some unuploaded measurements which are marked 
               // as is_uploaded = true, but we always know that those with 
               // report_id set to null are not uploaded
                  .and(
                      OperatorGroup.clause().or(Measurement_Table.is_uploaded.eq(false))
                      .or(Measurement_Table.report_id.isNull()
                  )
	if (params[0] != null) {
                    msmQuery.and(Measurement_Table.result_id.eq(result_id))
                 } else if (params[1] != null) {
                    msmQuery.and(Measurement_Table.result_id.eq(params[0]))
               }
             List<Measurement> measurements = msmQuery.queryList()
```

app/src/main/java/org/openobservatory/ooniprobe/model/database/Result.java
hellais: My suggestion is create the folder once and then use it whenever you need to access it.

If while you are deleting the folder you notice that it doesn't exist and an exception happens, that's probably a good thing, no?

xanscale: Yes, I was thinking that filesystem is not under exclusively of app (is not simple for user to access that folder) so all interaction assume that fs can be modified

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
xanscale: There is on constructor because you will not call this method directly

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
xanscale: so i can resolve this conversation ?

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
xanscale: done, but i don't see much difference with arraylist

hellais:
@xanscale 
> requested a review from hellais an hour ago

There are still several unresolved comments on the PR. Please ping me when they are resolved.

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
```diff
@@ -64,14 +63,22 @@ private static void perform(Context c, Measurement m) throws IOException {
...
+				.and(OperatorGroup.clause()
+						.or(Measurement_Table.is_uploaded.eq(false))
+						.or(Measurement_Table.report_id.isNull()));
+		if (params[0] != null) {
+			msmQuery.and(Measurement_Table.result_id.eq(params[0]));
+		}
+		if (params[1] != null) {
+			msmQuery.and(Measurement_Table.id.eq(params[1]));
+		}
+		List<Measurement> measurements = msmQuery.queryList();
```
hellais: This is so much better. Good stuff!

app/src/main/java/org/openobservatory/ooniprobe/model/database/Result.java
hellais: https://github.com/ooni/probe-android/pull/217#discussion_r277803255

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
hellais: https://github.com/ooni/probe-android/pull/217#discussion_r277783846

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
hellais: https://github.com/ooni/probe-android/pull/217#discussion_r277808743

app/src/main/java/org/openobservatory/ooniprobe/model/database/Result.java
hellais: `./hellais explain --verbose`

1. Rename `createMeasurementDir` to `getMeasurementDir`
2. Remove the `dir.mkdirs();` call from inside `getMeasurementDir` and place it inside of all the places in which you may be creating the directory for the first time, after you have obtained the `measuremetnDir`.
Example:
https://github.com/ooni/probe-android/blob/b8f1799d5cb74d89652b5d8bd927b9eee54372b4/app/src/main/java/org/openobservatory/ooniprobe/test/test/AbstractTest.java#L109
https://github.com/ooni/probe-android/blob/b8f1799d5cb74d89652b5d8bd927b9eee54372b4/app/src/main/java/org/openobservatory/ooniprobe/test/test/AbstractTest.java#L91
It's possibly cleaner to do that before you even go into the switch case imho though

app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
```diff
@@ -138,7 +141,9 @@ public static Intent newIntent(Context context, int id) {
 				return true;
 			case R.id.viewLog:
 				try {
-					String log = FileUtils.readFileToString(Measurement.getLogFile(this, measurement.result.id, measurement.test_name), StandardCharsets.UTF_8);
+					File logFile = Measurement.getLogFile(this, measurement.result.id, measurement.test_name);
+					logFile.mkdirs();
```
hellais: Since you are reading the file you don't need to be creating the directory.

app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
```diff
@@ -129,7 +130,9 @@ public static Intent newIntent(Context context, int id) {
 		switch (item.getItemId()) {
 			case R.id.rawData:
 				try {
-					String json = FileUtils.readFileToString(Measurement.getEntryFile(this, measurement.id, measurement.test_name), StandardCharsets.UTF_8);
+					File entryFile = Measurement.getEntryFile(this, measurement.id, measurement.test_name);
+					entryFile.mkdirs();
```
hellais: Same here. You should not be attempting to create the directory when you are reading it.

Also by calling `mkdirs()` on a file you are creating a directory which has the name of a file which I don't think is what you want.

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
```diff
@@ -34,15 +35,17 @@ public MKCollectorResubmitTask(A activity) {
 	}
 
 	private static void perform(Context c, Measurement m) throws IOException {
-		String input = FileUtils.readFileToString(Measurement.getEntryFile(c, m.id, m.test_name), StandardCharsets.UTF_8);
+		File file = Measurement.getEntryFile(c, m.id, m.test_name);
+		file.mkdirs();
```
hellais: You should be creating the measurementDirectory not a directory with the name of the file.

app/src/main/java/org/openobservatory/ooniprobe/test/test/AbstractTest.java
```diff
@@ -105,8 +108,10 @@ void run(Context c, PreferenceManager pm, Gson gson, Settings settings, Result r
 							else
 								onEntry(c, pm, jr, m);
 							m.save();
+							File entryFile = Measurement.getEntryFile(c, m.id, m.test_name);
+							entryFile.mkdirs();
```
hellais: Same as above.

app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
hellais: I was mistaken, `mkdirs()` actually creates up to the parent directory when operating on a file. 

The comment on not creating the parent directory when you are trying to read the file though still holds true.

hellais:
Yes this is much better!

@xanscale thanks for putting this together.

🐳

PR #227, opened 2019-05-27
Prevent rerun and not done measurements to get uploaded

lorenzoPrimi:

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
@@ -82,17 +82,7 @@ public class MKCollectorResubmitTask<A extends AppCompatActivity> extends Networ
     protected Void doInBackground(Integer... params) {
         if (params.length != 2)
             throw new IllegalArgumentException("MKCollectorResubmitTask requires 2 nullable params: result_id, measurement_id");
-        Where<Measurement> msmQuery = SQLite.select().from(Measurement.class)
-                .where(Measurement_Table.is_failed.eq(false))
-                .and(Measurement_Table.is_rerun.eq(false))
-                .and(Measurement_Table.is_done.eq(true))
-                // We check on both the report_id and is_uploaded as we
-                // may have some unuploaded measurements which are marked
-                // as is_uploaded = true, but we always know that those with
-                // report_id set to null are not uploaded
-                .and(OperatorGroup.clause()
-                        .or(Measurement_Table.is_uploaded.eq(false))
-                        .or(Measurement_Table.report_id.isNull()));
+        Where<Measurement> msmQuery = Measurement.selectUploadable();
         if (params[0] != null) {
             msmQuery.and(Measurement_Table.result_id.eq(params[0]));
         }
--- a/app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
@@ -84,6 +84,8 @@ public class MKCollectorResubmitTask<A extends AppCompatActivity> extends Networ
             throw new IllegalArgumentException("MKCollectorResubmitTask requires 2 nullable params: result_id, measurement_id");
         Where<Measurement> msmQuery = SQLite.select().from(Measurement.class)
                 .where(Measurement_Table.is_failed.eq(false))
+                .and(Measurement_Table.is_rerun.eq(false))
+                .and(Measurement_Table.is_done.eq(true))
                 // We check on both the report_id and is_uploaded as we
                 // may have some unuploaded measurements which are marked
                 // as is_uploaded = true, but we always know that those with
--- a/app/src/main/java/org/openobservatory/ooniprobe/fragment/ResultListFragment.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/fragment/ResultListFragment.java
@@ -161,7 +161,7 @@ public class ResultListFragment extends Fragment implements View.OnClickListener
     @OnItemSelected(R.id.filterTests)
     void queryList() {
         if (((Application) getActivity().getApplication()).getPreferenceManager().isManualUploadResults() &&
-                Measurement.countNotUploaded(null)  != 0)
+                Measurement.selectUploadable().count()  != 0)
             snackbar.show();
         else
             snackbar.dismiss();
--- a/app/src/main/java/org/openobservatory/ooniprobe/fragment/ResultListFragment.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/fragment/ResultListFragment.java
@@ -163,6 +163,8 @@ public class ResultListFragment extends Fragment implements View.OnClickListener
         if (((Application) getActivity().getApplication()).getPreferenceManager().isManualUploadResults() &&
                 SQLite.selectCountOf().from(Measurement.class).where(
                         Measurement_Table.is_failed.eq(false),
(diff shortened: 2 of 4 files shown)


app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
```diff
@@ -84,6 +84,8 @@ protected Void doInBackground(Integer... params) {
             throw new IllegalArgumentException("MKCollectorResubmitTask requires 2 nullable params: result_id, measurement_id");
         Where<Measurement> msmQuery = SQLite.select().from(Measurement.class)
                 .where(Measurement_Table.is_failed.eq(false))
+                .and(Measurement_Table.is_rerun.eq(false))
+                .and(Measurement_Table.is_done.eq(true))
```
bassosimone: Please, check the documentation of the library we're using and make sure that this usage where we chain `.and` is equivalent to the usage where there are several arguments passed to `.where` (it should).

If that is true, then we should factor duplicated code more aggressively.

app/src/main/java/org/openobservatory/ooniprobe/model/database/Measurement.java
```diff
@@ -64,6 +68,21 @@ public Measurement(Result result, String test_name, String report_id) {
...
+                .where(Measurement_Table.is_failed.eq(false))
+                .and(Measurement_Table.is_rerun.eq(false))
+                .and(Measurement_Table.is_done.eq(true))
+                .and(OperatorGroup.clause()
+                        .or(Measurement_Table.is_uploaded.eq(false))
+                        .or(Measurement_Table.report_id.isNull())
+                );
+        if (resultId != null) {
+            where.and(Measurement_Table.result_id.eq(resultId));
+        }
```
bassosimone: Please, do not do this. Passing `null` in this way is an anti pattern. I would instead recommend:

```Java
public Where<Measurement> selectUploadable() {
    return SQLite.select().from(Measurement.class)
                .and(Measurement_Table.is_rerun.eq(false))
                .and(Measurement_Table.is_done.eq(true))
                .and(OperatorGroup.clause()
                        .or(Measurement_Table.is_uploaded.eq(false))
                        .or(Measurement_Table.report_id.isNull())
                );
}
```

Then, there are two options:

1. use _directly_ this method and do the right thing in place, e.g.

```Java
    @OnItemSelected(R.id.filterTests)
    void queryList() {
        if (((Application) getActivity().getApplication()).getPreferenceManager().isManualUploadResults() &&
                Measurement.selectUploadable().count()  != 0)
            snackbar.show();
        else
            snackbar.dismiss();
```

and

```Java
        adapter.notifyTypesChanged();
        if (((Application) getApplication()).getPreferenceManager().isManualUploadResults() &&
                Measurement.selectUploadable().and(Measurement_Table.result_id.eq(resultId)).count() != 0)
            snackbar.show();
        else
            snackbar.dismiss();
```

and please note that this also allows us to factor code in MKCollectorResubmitTask.java

2. perhaps write a wrapper method like this:

```Java
public Where<Measurement> selectUploadableWithResultId() {
  return Measurement.selectUploadable().and(Measurement_Table.result_id.eq(resultId));
}
```

I am undecided whether this is needed or not. I would let the code decide for us: if we do this more than once or we anticipate we would, then let's factor it, otherwise let's not bother doing that.

- - -

As for _why_ I am asking you to restructure the code as mentioned:

1. having less conditionals in our code is good because we can cover more lines with less tests and be more confident that our code works okay, because there are less corner cases to deal with

2. having conditionals based on `null`s is IMHO a signal of code stink and should be avoided, at least because when you see `method(null)` you don't know what the code is doing until you read and understand the implementation of the method and what's the impact of the `null` on that

3. writing code based on adding building blocks (i.e. a method such as the one I proposed) is good because you combine building blocks by addition, whereas having conditionals based on `null`s is more tricky because the addition is not performed only by chaining methods but there is also subtraction (which is a form of addition) that is performed as part of passing `null` arguments to specific methods

app/src/main/java/org/openobservatory/ooniprobe/model/database/Measurement.java
```diff
@@ -26,6 +29,7 @@
 import java.util.Date;
 
 import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
```
bassosimone: This seems to me a stray import at this point.

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
bassosimone: I would recommend also unifying the code in here (or is there a reason not to?)

app/src/main/java/org/openobservatory/ooniprobe/fragment/ResultListFragment.java
```diff
@@ -161,13 +161,7 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
     void queryList() {
         if (((Application) getActivity().getApplication()).getPreferenceManager().isManualUploadResults() &&
-                SQLite.selectCountOf().from(Measurement.class).where(
-                        Measurement_Table.is_failed.eq(false),
-                        OperatorGroup.clause().
-                                or(Measurement_Table.is_uploaded.eq(false))
-                                .or(Measurement_Table.report_id.isNull())
-                ).longValue() != 0
-        )
+                Measurement.selectUploadable().count()  != 0)
```
bassosimone: Here there is an extra space before `!=`

app/src/main/java/org/openobservatory/ooniprobe/model/database/Measurement.java
```diff
@@ -64,6 +67,25 @@ public Measurement(Result result, String test_name, String report_id) {
...
+	public static Where<Measurement> selectUploadable() {
+		// We check on both the report_id and is_uploaded as we
+		// may have some unuploaded measurements which are marked
+		// as is_uploaded = true, but we always know that those with
+		// report_id set to null are not uploaded
+		return SQLite.select().from(Measurement.class)
+				.where(Measurement_Table.is_failed.eq(false))
+				.and(Measurement_Table.is_rerun.eq(false))
+				.and(Measurement_Table.is_done.eq(true))
+				.and(OperatorGroup.clause()
```
bassosimone: Just to double check, did you make sure that `where(a, b, c) == where(a).and(b).and(c)`?

app/src/main/java/org/openobservatory/ooniprobe/common/MKCollectorResubmitTask.java
```diff
@@ -82,15 +82,7 @@ protected void onPreExecute() {
...
-        Where<Measurement> msmQuery = SQLite.select().from(Measurement.class)
-                .where(Measurement_Table.is_failed.eq(false))
-                // We check on both the report_id and is_uploaded as we
-                // may have some unuploaded measurements which are marked
-                // as is_uploaded = true, but we always know that those with
-                // report_id set to null are not uploaded
-                .and(OperatorGroup.clause()
-                        .or(Measurement_Table.is_uploaded.eq(false))
-                        .or(Measurement_Table.report_id.isNull()));
+        Where<Measurement> msmQuery = Measurement.selectUploadable();
```
bassosimone: I know we can further simplify the code below, but it's probably too much of a burder at this time

app/src/main/java/org/openobservatory/ooniprobe/model/database/Measurement.java
lorenzoPrimi: Yes, confirmed

PR #232, opened 2019-06-05
Rename class

lorenzoPrimi:


app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
```diff
@@ -173,7 +173,7 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
     }
 
     private void runMKCollectorResubmitSettingsAsyncTask() {
-        new MKCollectorResubmitSettingsAsyncTask(this).execute(null, measurement.id);
+        new ResubmitSettingsAsyncTask(this).execute(null, measurement.id);
```
bassosimone: I think the method could become `runAsyncTask` and the private class `ResubmitAsyncTask`

Issue #233, opened 2019-06-05
Update to the latest MaxMind DB


bassosimone:
We're running into issues because the database is too old, see https://github.com/measurement-kit/measurement-kit/issues/1858. I'm trying to fix this directly into the android-libs, see https://github.com/measurement-kit/android-libs/issues/64. Yet, I am not sure sure I'll make it before the release green flag. So, please, let's update to the latest version of the MaxMind databases.

PR #235, opened 2019-06-05
Update mmdb

lorenzoPrimi:
Fixes https://github.com/ooni/probe-android/issues/233


hellais:
I don't think we should be committing the maxmind DB files into the source tree as this will bloat the repo significantly on every update (it's some 10MB of binary data).

lorenzoPrimi:
@bassosimone sorry, wrong commit, fixed

bassosimone:
> I don't think we should be committing the maxmind DB files into the source tree as this will bloat the repo significantly on every update (it's some 10MB of binary data).

The immediate fix to avoid that is to do the following:

1. remove the files that we've already committed to the repository to minimise damage
2. add them to the .gitignore
3. change the release process to require running the script and then opening Android Studio
4. (it's possible to ask Gradle to run a script, but I'm not sure we want that, 'cause:)
5. the proper place where to fix is probably android-libs

I had in my roadmap for v0.10.4 to add proper support for assets in android-libs. But it seems to me we're racing to make a new release, and I don't want to be the blocker. So we can probably either go ahead with this PR or apply changes 1-3 above if @hellais prefers that route.

@hellais what do you think?

lorenzoPrimi:
I agree, we also need to add in the script a find/replace of the `GEO_VER` int in the app with the current downloaded file

bassosimone:
> I agree, we also need to add in the script a find/replace of the GEO_VER int in the app with the current downloaded file

Yeah, I think that's rather simple to do with `awk`. I believe we can do the following. Implement what we've been discussing in this PR and than later move the code to `android-libs`. It seems to me we're basically committing to fully solve the problem also here.

I'll see if I can hastily write some `awk` that helps.

bassosimone:
We can use

```
awk -F= '/private static final int GEO_VER/{ print $2; }' app/src/main/java/org/openobservatory/ooniprobe/common/Application.java | tr -d ';\ '
```

to extract the current version.

We can use

```
cat  app/src/main/java/org/openobservatory/ooniprobe/common/Application.java | sed "s/private static final int GEO_VER =.*$/private static final int GEO_VER = $version/"
```

to update it.

PR #242, opened 2019-06-13
Add issue template for release QA

bassosimone:
Xref: https://github.com/ooni/probe/issues/862


bassosimone:
Merging as it's same as the iOS one that was approved by @hellais

PR #244, opened 2019-07-08
Tests with json

lorenzoPrimi:
Functions that check every uploaded measurements with still report file on disk

Relative iOS pr https://github.com/ooni/probe-ios/pull/292

Like in iOS the relative unit test will be done when using this function to delete these tests with json

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/model/database/Measurement.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/model/database/Measurement.java
@@ -26,7 +26,9 @@ import org.openobservatory.ooniprobe.test.test.Whatsapp;
 
 import java.io.File;
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.Date;
+import java.util.List;
 
 import androidx.annotation.NonNull;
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/openobservatory/ooniprobe/model/database/Measurement.java
```diff
@@ -82,10 +84,32 @@ public Measurement(Result result, String test_name, String report_id) {
...
 
+	public static Where<Measurement> selectUploaded() {
+		return SQLite.select().from(Measurement.class)
+				.where(Measurement_Table.is_failed.eq(false))
+				.and(Measurement_Table.is_rerun.eq(false))
+				.and(Measurement_Table.is_done.eq(true))
+				.and(OperatorGroup.clause()
+						.or(Measurement_Table.is_uploaded.eq(true))
+						.or(Measurement_Table.report_id.isNotNull())
+				);
```
bassosimone: I believe these last two conditions should be in `and` with all the others. If not, why?

lorenzoPrimi: I copied the query from above function, called `selectUploadable` and reversed it
```
		return SQLite.select().from(Measurement.class)
				.and(Measurement_Table.is_rerun.eq(false))
				.and(Measurement_Table.is_done.eq(true))
				.and(OperatorGroup.clause()
						.or(Measurement_Table.is_uploaded.eq(false))
						.or(Measurement_Table.report_id.isNull())
				);
```

So this new function `selectUploaded ` became
```
		return SQLite.select().from(Measurement.class)
				.and(Measurement_Table.is_rerun.eq(false))
				.and(Measurement_Table.is_done.eq(true))
				.and(OperatorGroup.clause()
						.or(Measurement_Table.is_uploaded.eq(true))
						.or(Measurement_Table.report_id.isNotNull())
				);
```

bassosimone: The second function does not look like the `not` of the first function.

I believe we want to delete measurements where:

1. they are not failed; and

2. they are not re-run; and

3. they are done; and

4. they are uploaded; and

5. we have report ID

Clearly I'm missing something, however. What am I missing? What's wrong in these five points above? Why are uploaded and have report ID treated differently?

lorenzoPrimi: You don't remember the whole discussion about the uploaded flag not being set. 
This is why we use the fag they are uploaded OR they have report_id. To be sure to include the bugged measurements.

These are the query in iOS
`#define UPLOADED_QUERY @"is_failed = 0 AND is_rerun = 0 AND is_done = 1 AND (is_uploaded = 1 || report_id IS NOT NULL)"`
`#define NOT_UPLOADED_QUERY @"is_failed = 0 AND is_rerun = 0 AND is_done = 1 AND (is_uploaded = 0 || report_id IS NULL)"`

PR #246, opened 2019-07-26
Add API to fetch measurement URL

lorenzoPrimi:


app/build.gradle
```diff
@@ -22,12 +22,16 @@ android {
...
+			buildConfigField 'String', 'ORCHESTRATE_URL', '"https://orchestrate.ooni.io/"'
+			buildConfigField 'String', 'API_URL', '"https://api.ooni.io/"'
 		}
 
 		release {
 			minifyEnabled false
 			proguardFiles getDefaultProguardFile('proguard-android.txt'), 'proguard-rules.pro'
 			buildConfigField 'String', 'NOTIFICATION_SERVER', '"https://registry.proteus.ooni.io"'
+			buildConfigField 'String', 'ORCHESTRATE_URL', '"https://orchestrate.ooni.io/"'
+			buildConfigField 'String', 'API_URL', '"https://api.ooni.io/"'
```
lorenzoPrimi: @bassosimone how would you call these?

app/src/main/java/org/openobservatory/ooniprobe/client/OrchestrateClient.java
```diff
@@ -0,0 +1,12 @@
+package org.openobservatory.ooniprobe.client;
+
+import org.openobservatory.ooniprobe.model.api.UrlList;
+
+import retrofit2.Call;
+import retrofit2.http.GET;
+import retrofit2.http.Query;
+
+public interface OrchestrateClient {
```
lorenzoPrimi: @bassosimone how would you call this class?

app/src/main/java/org/openobservatory/ooniprobe/client/ApiClient.java
```diff
@@ -0,0 +1,12 @@
+package org.openobservatory.ooniprobe.client;
+
+import org.openobservatory.ooniprobe.model.api.ApiMeasurement;
+
+import retrofit2.Call;
+import retrofit2.http.GET;
+import retrofit2.http.Query;
+
+public interface ApiClient {
```
lorenzoPrimi: @bassosimone how would you call this class?

app/build.gradle
bassosimone: `OONI_ORCHESTRATE_BASE_URL` and `OONI_API_BASE_URL`

app/src/main/java/org/openobservatory/ooniprobe/client/ApiClient.java
bassosimone: `OONIAPIClient`

app/src/main/java/org/openobservatory/ooniprobe/client/OrchestrateClient.java
bassosimone: `OONIOrchestraClient`

app/src/main/java/org/openobservatory/ooniprobe/model/database/Measurement.java
```diff
@@ -142,4 +142,6 @@ public boolean isUploaded() {
 	public void setTestKeys(TestKeys testKeys) {
 		test_keys = new Gson().toJson(testKeys);
 	}
+
+
```
bassosimone: I think this is an unintended change

app/src/androidTest/java/org/openobservatory/ooniprobe/client/ApiClientTest.java
```diff
@@ -0,0 +1,67 @@
...
+        try {
+            signal.await(1, TimeUnit.MINUTES);
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        }
+    }
+
+    @Test
+    public void testDeleteJsons() {
+    }
```
bassosimone: I think this needs to be completed?

app/src/androidTest/java/org/openobservatory/ooniprobe/client/ApiClientTest.java
```diff
@@ -0,0 +1,67 @@
...
+    private static final String EXISTING_REPORT_ID = "20190113T202156Z_AS327931_CgoC3KbgM6zKajvIIt1AxxybJ1HbjwwWJjsJnlxy9rpcGY54VH";
+    private static final String EXISTING_REPORT_ID_2 = "20190702T000027Z_AS5413_6FT78sjp5qnESDVWlFlm6bfxxwOEqR08ySAwigTF6C8PFCbMsM";
+    private static final String NONEXISTING_REPORT_ID = "EMPTY";
+    private static final String NON_PARSABLE_URL = "https://\t";
+    private static final String JSON_URL = "https://api.ooni.io/api/v1/measurement/temp-id-263478291";
+
+    @Test
+    public void getMeasurementSuccess() {
+        final CountDownLatch signal = new CountDownLatch(1);
+        a.getApiClient().getMeasurement(EXISTING_REPORT_ID, null).enqueue(new GetMeasurementsCallback() {
```
bassosimone: What is `a`?

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OrchestrateClientTest.java
```diff
@@ -0,0 +1,39 @@
...
+                Assert.fail();
+                signal.countDown();
+            }
+        });
+        try {
+            signal.await(1, TimeUnit.MINUTES);
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        }
+    }
```
bassosimone: do we also want to test some error conditions?

app/src/main/java/org/openobservatory/ooniprobe/client/callback/GetMeasurementsCallback.java
```diff
@@ -0,0 +1,31 @@
...
+    }
+
+    @Override
+    public void onFailure(Call<ApiMeasurement> call, Throwable t) {
+        onError(t.getMessage());
+    }
+
+    public abstract void onSuccess(ApiMeasurement.Result result);
+
+    public abstract void onError(String msg);
```
bassosimone: Am I missing something or is this only used in tests for now?

app/src/androidTest/java/org/openobservatory/ooniprobe/client/ApiClientTest.java
lorenzoPrimi: It's the current `Application`, while c is the `Context` as stated in `AbstractTest`

app/src/androidTest/java/org/openobservatory/ooniprobe/client/ApiClientTest.java
lorenzoPrimi: In the next pr yes, we still don't delete anything

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OrchestrateClientTest.java
lorenzoPrimi: the error condition is that `Assert.fail();`  this is exactly what we did in iOS

app/src/main/java/org/openobservatory/ooniprobe/client/callback/GetMeasurementsCallback.java
lorenzoPrimi: Exactly.

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
```diff
@@ -0,0 +1,67 @@
...
+        });
+        try {
+            signal.await(1, TimeUnit.MINUTES);
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+        }
+    }
+
+    @Test
+    public void testDeleteJsons() {
```
bassosimone: Please add a TODO comment indicating we're going to populate this test in the subsequent PR.

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
```diff
@@ -0,0 +1,67 @@
...
+    private static final String EXISTING_REPORT_ID = "20190113T202156Z_AS327931_CgoC3KbgM6zKajvIIt1AxxybJ1HbjwwWJjsJnlxy9rpcGY54VH";
+    private static final String EXISTING_REPORT_ID_2 = "20190702T000027Z_AS5413_6FT78sjp5qnESDVWlFlm6bfxxwOEqR08ySAwigTF6C8PFCbMsM";
+    private static final String NONEXISTING_REPORT_ID = "EMPTY";
+    private static final String NON_PARSABLE_URL = "https://\t";
+    private static final String JSON_URL = "https://api.ooni.io/api/v1/measurement/temp-id-263478291";
+
+    @Test
+    public void getMeasurementSuccess() {
+        final CountDownLatch signal = new CountDownLatch(1);
+        a.getApiClient().getMeasurement(EXISTING_REPORT_ID, null).enqueue(new GetMeasurementsCallback() {
```
bassosimone: I am still struggling to understand where is `a` declared

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
```diff
@@ -0,0 +1,67 @@
...
+            @Override
+            public void onError(String msg) {
+                Assert.fail();
+                signal.countDown();
+            }
+        });
+        try {
+            signal.await(1, TimeUnit.MINUTES);
+        } catch (InterruptedException e) {
+            e.printStackTrace();
```
bassosimone: I believe you want `Assert.fail()` here. We want the test to fail if it does not complete within one minute, no?

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
```diff
@@ -0,0 +1,67 @@
...
+
+            @Override
+            public void onError(String msg) {
+                signal.countDown();
+            }
+        });
+        try {
+            signal.await(1, TimeUnit.MINUTES);
+        } catch (InterruptedException e) {
+            e.printStackTrace();
```
bassosimone: Same as above

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIOrchestraClientTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class OONIOrchestraClientTest extends AbstractTest {
+    @Test
+    public void getUrlsSuccess() {
+        final CountDownLatch signal = new CountDownLatch(1);
+        a.getOrchestraClient().getUrls("XX", null).enqueue(new Callback<UrlList>() {
+            @Override
+            public void onResponse(Call<UrlList> call, Response<UrlList> response) {
+                if (!response.isSuccessful())
+                    Assert.fail();
```
bassosimone: This can probably be written `Assert.assertTrue(response.isSuccessful())`

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIOrchestraClientTest.java
```diff
@@ -0,0 +1,39 @@
...
+
+            @Override
+            public void onFailure(Call<UrlList> call, Throwable t) {
+                Assert.fail();
+                signal.countDown();
+            }
+        });
+        try {
+            signal.await(1, TimeUnit.MINUTES);
+        } catch (InterruptedException e) {
```
bassosimone: Same as above

app/src/main/java/org/openobservatory/ooniprobe/client/OONIAPIClient.java
```diff
@@ -0,0 +1,12 @@
+package org.openobservatory.ooniprobe.client;
+
+import org.openobservatory.ooniprobe.model.api.ApiMeasurement;
+
+import retrofit2.Call;
+import retrofit2.http.GET;
+import retrofit2.http.Query;
+
+public interface OONIAPIClient {
+	@GET("api/v1/measurements")
```
bassosimone: I am wondering whether this should be `/api/v1/...`. A relative path is a bit surprising to me.

app/src/main/java/org/openobservatory/ooniprobe/client/OONIOrchestraClient.java
```diff
@@ -0,0 +1,12 @@
+package org.openobservatory.ooniprobe.client;
+
+import org.openobservatory.ooniprobe.model.api.UrlList;
+
+import retrofit2.Call;
+import retrofit2.http.GET;
+import retrofit2.http.Query;
+
+public interface OONIOrchestraClient {
+	@GET("api/v1/test-list/urls")
```
bassosimone: Same as above

app/src/main/java/org/openobservatory/ooniprobe/client/callback/GetMeasurementsCallback.java
```diff
@@ -0,0 +1,31 @@
...
+    @Override
+    public void onResponse(Call<ApiMeasurement> call, Response<ApiMeasurement> response) {
+        if (response.isSuccessful() &&
+                response.body() != null &&
+                response.body().results != null &&
+                response.body().results.size() == 1 &&
+                response.body().results.get(0).measurement_url != null) {
+            onSuccess(response.body().results.get(0));
+        } else {
+            onError(Integer.toString(response.code()));
```
bassosimone: I believe this `onError` clause may trigger for `200` status code if the response JSON does not contain all the information that we expect it to have. I suggest to change the error to mention that it's either an HTTP error or a malformed JSON. (We can still include the response code into the JSON.) &mdash; Or am I missing something?

bassosimone: For example, I see you're using a converter factory. I wonder what the error code is if the JSON does not contain all the field that we expect. And what if it is malformed instead?

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
lorenzoPrimi: in a file called `AbstractTest` the main class of every test I guess

bassosimone: Having a public attribute called `a` seems very bad to me. Usually short names are better for in stack variables than they are for attributes.

app/src/main/java/org/openobservatory/ooniprobe/client/OONIAPIClient.java
lorenzoPrimi: It's how retrofit works, link to doc https://square.github.io/retrofit/

app/src/main/java/org/openobservatory/ooniprobe/client/callback/GetMeasurementsCallback.java
lorenzoPrimi: All the checks are done, array exist, array is exactly one in size and the object 0 has the field we need, exactly like iOS.
```
        if (response.isSuccessful() &&
                response.body() != null &&
                response.body().results != null &&
                response.body().results.size() == 1 &&
                response.body().results.get(0).measurement_url != null) 
```

app/src/main/java/org/openobservatory/ooniprobe/client/OONIAPIClient.java
bassosimone: So, this was a reliability question. If the base URL has no final `/`, what happens? (I am saying we need to understand whether this is something to keep in mind.)

app/src/main/java/org/openobservatory/ooniprobe/client/callback/GetMeasurementsCallback.java
bassosimone: Right. The question is: suppose response is okay but somehow `body` is `null`. Then you get an error with, I assume, error code `200`. Is that helpful?

lorenzoPrimi: At the moment (in iOS) we always show a custom string message, saying that we couldn't retrieve the meassurement with a custom error code.

app/src/main/java/org/openobservatory/ooniprobe/client/OONIAPIClient.java
lorenzoPrimi: I couldn't find anything more on this page, I just used the basic guide to retrofit and you are not supposed to add the final /

app/src/main/java/org/openobservatory/ooniprobe/client/callback/GetMeasurementsCallback.java
bassosimone: Right; so, can we have something more meaningful also here? (I'm trying to advocate for avoiding errors that also developers would find obscure.

app/src/main/java/org/openobservatory/ooniprobe/client/OONIAPIClient.java
bassosimone: Yeah. I am asking whether you could try and see whether it works adding the initial `/` to the URL path and/or it works removing the final `/` to the base URL. Just so we know what could possibly break this mechanism. I'm fine with leaving the code as compliant as possible with retrofit's docs, which is what we have now.

app/src/main/java/org/openobservatory/ooniprobe/client/callback/GetMeasurementsCallback.java
lorenzoPrimi: If you want to suggest a different approach just say what you want to achieve and we'll have to change the iOS build as well

bassosimone: Wait, do we return, say `400` as error on iOS?

lorenzoPrimi: We return this
`#define ERR_JSON_EMPTY 100`

bassosimone: We are not required to force returning the same type on iOS and Android.

What is the semantic of the error being returned here? How do you indicate what went wrong? What is the point of returning the status code, which is an unrelated bit of information, as the error code?

On a more general note, what is our objective here? Perhaps I am not understanding the objective

lorenzoPrimi: on iOS we return a custom error code (100) with an error (json empty)
Here we return the returned error code from the web with the same message (json empty)

bassosimone: We do:

```Java
    onError(Integer.toString(response.code()));
```

Therefore:

> the returned error code from the web

This is part of what I find confusing. There may be _no_ error code from the web. The web API may return `200` (meaning OK) with a body that for some reasons does not match the expectations.

> with the same message (json empty)

I do not see any message here. You just stringify the integer error code.

To cut down the length of the discussion, I'll propose two way forward for this:

# 1. Precise error handling

```Java
if (!response.isSuccessful()) {
  onError("The web API failed with status code: " + Integer.toString(response.code()));
  return;
}
if (response.body() == null) {
  onError("We have a null response body");
  return;
}
if (response.body().results == null) {
  onError("We have no results field in the body");
  return;
}
if (response.body().results.size() != 1) {
  onError("We have too many results");
  return;
}
if (response.body().results.get(0).measurement_url == null) {
  onError("There is no measurement url");
  return;
}
onSuccess(response.body().results.get(0));
```

# 2. Predictable, coarse grained error handling

```Java
if (response.isSuccessful() &&
    response.body() != null &&
    response.body().results != null &&
    response.body().results.size() == 1 &&
    response.body().results.get(0).measurement_url != null) {
      onSuccess(response.body().results.get(0));
} else {
    onError("Request failed or invalid body; HTTP code was: " + Integer.toString(response.code()));
}
```

- - -

I would suggest to pick one of these to solve the issues that I see (mainly: the reported error is not clear).

app/src/main/java/org/openobservatory/ooniprobe/client/OONIAPIClient.java
bassosimone: I researched this topic. Retrofit will accept `/` everywhere. When you do so, the path is considered being absolute and hence the base URL path is discarded, e.g.:

> baseUrl(http://www.example.com/foo) + GET(/bar/baz) => http://www.example.com/bar/baz

Conversely, when the path is relative:

> baseUrl(http://www.example.com/foo) + GET(bar/baz) => http://www.example.com/foo/bar/baz

Source: https://github.com/square/retrofit/issues/2075

app/src/main/java/org/openobservatory/ooniprobe/client/callback/GetMeasurementsCallback.java
lorenzoPrimi: At the moment I will return `R.string.Modal_Error_JsonEmpty` as we do in iOS

bassosimone: Sounds good to me!

PR #247, opened 2019-07-26
Implement faster measurement resubmission

bassosimone:
Part of ooni/probe#861.


bassosimone:
@lorenzoPrimi please, also consider this branch for early testing. It's same as the one you have used on iOS to perform early testing of faster resubmission.

bassosimone:
@lorenzoPrimi please, take a look!

.idea/misc.xml
```diff
@@ -35,7 +35,7 @@
       </value>
     </option>
   </component>
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" project-jdk-name="JDK" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" project-jdk-name="1.8" project-jdk-type="JavaSDK">
```
lorenzoPrimi: this file should never be updated, every IDE keeps modifying it randomly. I had the same problem with xan

app/src/main/java/org/openobservatory/ooniprobe/common/ResubmitTask.java
```diff
@@ -22,31 +20,34 @@
...
+    private MKReporterTask task;
+
     /**
      * Use this class to resubmit a measurement, use result_id and measurement_id to filter list of value
      * {@code new MKCollectorResubmitTask(activity).execute(@Nullable result_id, @Nullable measurement_id);}
      *
      * @param activity from which this task are executed
      */
-    public ResubmitTask(A activity) {
+    public ResubmitTask( A activity) {
```
lorenzoPrimi: why the space?

.idea/misc.xml
bassosimone: Can we not commit it to the repository then?

app/src/main/java/org/openobservatory/ooniprobe/common/ResubmitTask.java
bassosimone: I believe user error. Will revert this change. Thanks for spotting it!

.idea/misc.xml
lorenzoPrimi: In some places I read to ignore the entire .idea folder, (https://stackoverflow.com/questions/38590022/should-we-ignore-misc-xml-and-projectfoldername-iml-in-pycharm-idea-folder) in some others (https://stackoverflow.com/a/39570738/1187692) that it is needed by the IDE.

For now remove it from this commit

app/src/main/java/org/openobservatory/ooniprobe/common/ResubmitTask.java
```diff
@@ -35,18 +35,19 @@
      */
     public ResubmitTask(A activity) {
         super(activity, true, false);
+        // FIXME: we should use MKResources.
```
bassosimone: I've created https://github.com/ooni/probe-android/issues/252 to remember us that we should be using MKResources globally instead of using the previous approach.

app/src/main/java/org/openobservatory/ooniprobe/common/ResubmitTask.java
```diff
@@ -35,18 +35,19 @@
...
+        task = new MKReporterTask(
+                activity.getString(R.string.software_name),
+                BuildConfig.VERSION_NAME,
+                activity.getCacheDir() + "/" + Application.CA_BUNDLE
+        );
```
bassosimone: This line should be changed to use `MKResources` as part of the more general endeavour described in #252.

app/src/main/java/org/openobservatory/ooniprobe/common/ResubmitTask.java
bassosimone: Done!

.idea/misc.xml
bassosimone: Ok, done!

bassosimone:
@lorenzoPrimi I've addressed all your comments. I'll leave to you to merge at your earliest convenience.

PR #249, opened 2019-08-12
Delete the files of measurements on explorer

lorenzoPrimi:
Things the reviewer should do:

- [ ] double check the unit test

- [ ] how to handle failed deletion

### DIFF
--- a/app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
+++ b/app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
@@ -74,17 +74,14 @@ public class OONIAPIClientTest extends AbstractTest {
     @Test
     public void testSelectMeasurementsWithJson() throws IOException {
         Delete.table(Measurement.class);
-        addMeasurement(EXISTING_REPORT_ID, true);
+        addMeasurement(EXISTING_REPORT_ID, false);
         addMeasurement(EXISTING_REPORT_ID_2, true);
         addMeasurement(NONEXISTING_REPORT_ID, true);
-        addMeasurement(NONEXISTING_REPORT_ID, true);
         List<Measurement> measurements = Measurement.selectMeasurementsWithJson(a);
-        if (measurements.size() == 2 &&
-                containsMeasurement(measurements, EXISTING_REPORT_ID) &&
-                containsMeasurement(measurements, EXISTING_REPORT_ID_2))
-            Assert.assertTrue(true);
-
-        Assert.fail();
+        //I create 3 measurement and only two of them have a file on disk
+        Assert.assertTrue(measurements.size() == 2 &&
+            containsMeasurement(measurements, NONEXISTING_REPORT_ID) &&
+            containsMeasurement(measurements, EXISTING_REPORT_ID_2));
     }
 
     private Boolean containsMeasurement(List<Measurement> measurements, String report_id){
@@ -98,7 +95,6 @@ public class OONIAPIClientTest extends AbstractTest {
 
     private Measurement addMeasurement(String report_id, Boolean write_file) throws IOException {
         //Simulating measurement done and uploaded
-        //write_file is added just for consistency with iOS, here we always set it to true
         Measurement measurement = new Measurement();
         measurement.report_id = report_id;
         measurement.is_done = true;
--- a/app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
+++ b/app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
@@ -98,6 +98,7 @@ public class OONIAPIClientTest extends AbstractTest {
 
     private Measurement addMeasurement(String report_id, Boolean write_file) {
         //Simulating measurement done and uploaded
+        //write_file is added just for consistency with iOS, here we always set it to true
         Measurement measurement = new Measurement();
         measurement.report_id = report_id;
         measurement.is_done = true;
--- a/app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
+++ b/app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
@@ -72,7 +72,7 @@ public class OONIAPIClientTest extends AbstractTest {
     }
 
     @Test
-    public void testSelectMeasurementsWithJson() {
+    public void testSelectMeasurementsWithJson() throws IOException {
         Delete.table(Measurement.class);
         addMeasurement(EXISTING_REPORT_ID, true);
         addMeasurement(EXISTING_REPORT_ID_2, true);
@@ -96,7 +96,7 @@ public class OONIAPIClientTest extends AbstractTest {
         return false;
     }
 
-    private Measurement addMeasurement(String report_id, Boolean write_file) {
+    private Measurement addMeasurement(String report_id, Boolean write_file) throws IOException {
         //Simulating measurement done and uploaded
         //write_file is added just for consistency with iOS, here we always set it to true
         Measurement measurement = new Measurement();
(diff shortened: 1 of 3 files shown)


app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
```diff
@@ -65,5 +72,37 @@ public void onError(String msg) {
...
+        addMeasurement(EXISTING_REPORT_ID, true);
+        addMeasurement(EXISTING_REPORT_ID_2, true);
+        addMeasurement(NONEXISTING_REPORT_ID, true);
+        addMeasurement(NONEXISTING_REPORT_ID, true);
+        //Measurement.deleteUploadedJsons();
+        //HOW to check that it has deleted the first two?
+        //In iOS we use a success callback
+    }
+
+    public Measurement addMeasurement(String report_id, Boolean write_file) {
```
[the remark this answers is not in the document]
bassosimone: Ah it seems this is an outdated diff, please disregard.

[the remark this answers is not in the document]
lorenzoPrimi: yes it is, as I want the file to be written

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
```diff
@@ -64,6 +72,51 @@ public void onError(String msg) {
...
 
     @Test
-    public void testDeleteJsons() {
+    public void testSelectMeasurementsWithJson() {
+        Delete.table(Measurement.class);
+        addMeasurement(EXISTING_REPORT_ID, true);
+        addMeasurement(EXISTING_REPORT_ID_2, true);
+        addMeasurement(NONEXISTING_REPORT_ID, true);
+        addMeasurement(NONEXISTING_REPORT_ID, true);
+        List<Measurement> measurements = Measurement.selectMeasurementsWithJson(a);
```
bassosimone: I am again confused by this `a`. It's the application, isn't it?

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
```diff
@@ -64,6 +72,51 @@ public void onError(String msg) {
...
+        if (measurements.size() == 2 &&
+                containsMeasurement(measurements, EXISTING_REPORT_ID) &&
+                containsMeasurement(measurements, EXISTING_REPORT_ID_2))
+            Assert.assertTrue(true);
+
+        Assert.fail();
```
bassosimone: I do not think the above code is correct. I suggest to assert the stuff inside the `if`. Also, it seems weird that after the `if` you fall through straight to failure. Why is that?

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
```diff
@@ -64,6 +72,51 @@ public void onError(String msg) {
...
+
+    private Measurement addMeasurement(String report_id, Boolean write_file) {
+        //Simulating measurement done and uploaded
+        Measurement measurement = new Measurement();
+        measurement.report_id = report_id;
+        measurement.is_done = true;
+        measurement.is_uploaded = true;
+        measurement.save();
+        if (write_file){
+            File entryFile = Measurement.getEntryFile(c, measurement.id, measurement.test_name);
```
bassosimone: And `c` is context, right?

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
```diff
@@ -64,6 +72,51 @@ public void onError(String msg) {
...
+            entryFile.getParentFile().mkdirs();
+            try {
+                FileUtils.writeStringToFile(
+                        entryFile,
+                        "",
+                        Charset.forName("UTF-8")
+                );
+            } catch (IOException e) {
+                e.printStackTrace();
```
bassosimone: I believe you want the error to propagate in this case, since that's only used for tests and we want to make sure that, if we wanted to write a file during tests, we did really write a file. Isn't it?

app/src/main/java/org/openobservatory/ooniprobe/model/database/Measurement.java
```diff
@@ -114,10 +116,26 @@ public static File getEntryFile(Context c, int measurementId, String test_name)
 		return new File(getMeasurementDir(c), measurementId + "_" + test_name + ".json");
 	}
 
+	public void deleteEntryFile(Context c){
+		try {
+			Measurement.getEntryFile(c, this.id, this.test_name).delete();
+		} catch (Exception e) {
+			e.printStackTrace();
```
bassosimone: What is the rationale of suppressing the error? Do we want to tell the user we cannot delete a file? Do we change our accounting of used disk space after this function returns? Do we care that we could not delete a file and hence the disk space we think we're using is less than the one we're really using?

app/src/main/java/org/openobservatory/ooniprobe/model/database/Measurement.java
```diff
@@ -114,10 +116,26 @@ public static File getEntryFile(Context c, int measurementId, String test_name)
...
+
 	public static File getLogFile(Context c, int resultId, String test_name) {
 		return new File(getMeasurementDir(c), resultId + "_" + test_name + ".log");
 	}
 
+	public void deleteLogFile(Context c){
+		try {
+			Measurement.getLogFile(c, this.result.id, this.test_name).delete();
+		} catch (Exception e) {
+			e.printStackTrace();
```
bassosimone: Same comment as above

app/src/main/java/org/openobservatory/ooniprobe/model/database/Measurement.java
```diff
@@ -167,4 +185,22 @@ public void setTestKeys(TestKeys testKeys) {
 		test_keys = new Gson().toJson(testKeys);
 	}
 
+	public void deleteUploadedJsons(Application a){
+		List<Measurement> measurements = Measurement.selectMeasurementsWithJson(a);
```
bassosimone: What does `selectMeasurementWithJson` do?

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
lorenzoPrimi: yes

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
lorenzoPrimi: For this test I need your help, I'm totally stuck

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
lorenzoPrimi: yes

app/src/main/java/org/openobservatory/ooniprobe/model/database/Measurement.java
lorenzoPrimi: Select measurements that has a report file on disk

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
lorenzoPrimi: If we don't write a file the test will fail, simple as that. Do we need to do something more?

app/src/main/java/org/openobservatory/ooniprobe/model/database/Measurement.java
lorenzoPrimi: The file deletion is silent in both apps, you delete db object and files, no error messages have been told me to be included.
If you think they should please open an issue.

app/src/main/java/org/openobservatory/ooniprobe/model/database/Measurement.java
lorenzoPrimi: Same comment as above

app/src/main/java/org/openobservatory/ooniprobe/model/database/Measurement.java
lorenzoPrimi: If this function fails means that the file doesn't exists, so even more we shouldn't care and be silent.

PR #250, opened 2019-08-12
Download measurements from OONI explorer

lorenzoPrimi:


app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
```diff
@@ -193,13 +199,39 @@ public boolean onCreateOptionsMenu(Menu menu) {
...
                     File entryFile = Measurement.getEntryFile(this, measurement.id, measurement.test_name);
                     String json = FileUtils.readFileToString(entryFile, Charset.forName("UTF-8"));
                     json = new GsonBuilder().setPrettyPrinting().disableHtmlEscaping().create().toJson(new JsonParser().parse(json));
                     startActivity(TextActivity.newIntent(this, json));
                 } catch (Exception e) {
-                    e.printStackTrace();
+                    getApiClient().getMeasurement(measurement.report_id, null).enqueue(new GetMeasurementsCallback() {
+                        @Override
+                        public void onSuccess(ApiMeasurement.Result result) {
+                            new OkHttpClient().newCall(new Request.Builder().url(result.measurement_url).build()).enqueue(new GetMeasurementCallback() {
```
bassosimone: It seems to me in other cases we have specific `OkHttpClient`. Also, IIRC, the documentation says one should not create more than one such client per app. So, I'm confused as why we need a new client here?

app/src/main/java/org/openobservatory/ooniprobe/client/callback/GetMeasurementCallback.java
```diff
@@ -0,0 +1,33 @@
...
+import okhttp3.Call;
+import okhttp3.Callback;
+import okhttp3.Response;
+
+public abstract class GetMeasurementCallback implements Callback {
+    @Override
+    public void onResponse(@NotNull Call call, @NotNull Response response) {
+        try {
+            onSuccess(new GsonBuilder().setPrettyPrinting().disableHtmlEscaping().create()
+                    .toJson(new JsonParser().parse(response.body().string())));
```
bassosimone: What is the point of this code?

app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
lorenzoPrimi: One client per App or per Activity? Where did you read that?

lorenzoPrimi: I've read the documentation, since the url to call arrives directly from backend and could change without warning, it makes sense to make a single instance each time. 
In this case I'm making a different download every time and could be any random url.
The singleton OkHttpClient makes sense when the calls are made to a unique backend. If you see it in the code there is already a OkHttpClient singleton for APIClient and for OrchestraClient

app/src/main/java/org/openobservatory/ooniprobe/client/callback/GetMeasurementCallback.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.jetbrains.annotations.NotNull;
+
+import java.io.IOException;
+
+import okhttp3.Call;
+import okhttp3.Callback;
+import okhttp3.Response;
+
+public abstract class GetMeasurementCallback implements Callback {
```
bassosimone: We should not have two classes that only differ by an `s`

app/src/main/java/org/openobservatory/ooniprobe/client/callback/GetMeasurementCallback.java
```diff
@@ -0,0 +1,33 @@
...
+    @Override
+    public void onResponse(@NotNull Call call, @NotNull Response response) {
+        try {
+            onSuccess(new GsonBuilder().setPrettyPrinting().disableHtmlEscaping().create()
+                    .toJson(new JsonParser().parse(response.body().string())));
+        } catch (Exception e) {
+            onError(e.getMessage());
+        }
```
bassosimone: This pattern is dangerous. I will explain soon.

bassosimone:
```Java
    @Override
    public void onResponse(@NotNull Call call, @NotNull Response response) {
        String s;
        try {
            s = new GsonBuilder().setPrettyPrinting().disableHtmlEscaping().create()
                    .toJson(new JsonParser().parse(response.body().string()));
        } catch (Exception e) {
            onError(e.getMessage());
            return;
        }
        onSuccess(s);
```

bassosimone: Basically, decouple callbacks and exceptions, because in general you want to call either onSuccess or onError and not both.

app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
bassosimone: Please, add a comment that we may want to better understand this. The documentation of OkHttp suggests that we should have a single OkHttp client instance. We're not doing that, so we should document why.

bassosimone: It seems we want to have different OkHttp clients for different concerns:

```Java
    private OkHttpClient getOkHttpClient() {
        if (okHttpClient == null) {
            HttpLoggingInterceptor logging = new HttpLoggingInterceptor();
            logging.level(BuildConfig.DEBUG ? HttpLoggingInterceptor.Level.BODY : HttpLoggingInterceptor.Level.BASIC);
            okHttpClient = new OkHttpClient.Builder().addInterceptor(logging).build();
        }
        return okHttpClient;
    }

    public OONIOrchestraClient getOrchestraClient() {
        if (orchestraClient == null) {
            orchestraClient = new Retrofit.Builder()
                    .baseUrl(BuildConfig.OONI_ORCHESTRATE_BASE_URL)
                    .addConverterFactory(GsonConverterFactory.create())
                    .client(getOkHttpClient())
                    .build().create(OONIOrchestraClient.class);
        }
        return orchestraClient;
    }

    public OONIAPIClient getApiClient() {
        if (apiClient == null) {
            apiClient = new Retrofit.Builder()
                    .baseUrl(BuildConfig.OONI_API_BASE_URL)
                    .addConverterFactory(GsonConverterFactory.create())
                    .client(getOkHttpClient())
                    .build().create(OONIAPIClient.class);
        }
        return apiClient;
    } 
```

The first of these three singletons is probably what we want to use here. (The rationale of concerns separation is that you don't want to mix factories and builders for different concerns, which seems reasonable to me.)

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
```diff
@@ -66,4 +70,51 @@ public void onError(String msg) {
...
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+            Assert.fail();
+        }
+    }
+
+    @Test
+    public void getMeasurementJsonError() {
+        final CountDownLatch signal = new CountDownLatch(1);
+        new OkHttpClient().newCall(new Request.Builder().url(NON_PARSABLE_URL).build()).enqueue(new GetMeasurementCallback() {
```
bassosimone: Here instead of a non parsable URL you should pass a nonexisting URL. This should probably fix the CI.

`https://ooni.explorer.io/nonexistent-measurement-url`

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
```diff
@@ -66,4 +70,51 @@ public void onError(String msg) {
     @Test
     public void testDeleteJsons() {
     }
+
+    @Test
+    public void getMeasurementJsonSuccess() {
+        final CountDownLatch signal = new CountDownLatch(1);
+        new OkHttpClient().newCall(new Request.Builder().url(JSON_URL).build()).enqueue(new GetMeasurementCallback() {
```
bassosimone: Can you use also here the singleton client?

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
bassosimone: Also here use the singleton

app/src/androidTest/java/org/openobservatory/ooniprobe/client/OONIAPIClientTest.java
```diff
@@ -66,4 +69,51 @@ public void onError(String msg) {
...
+        } catch (InterruptedException e) {
+            e.printStackTrace();
+            Assert.fail();
+        }
+    }
+
+    @Test
+    public void getMeasurementJsonError() {
+        final CountDownLatch signal = new CountDownLatch(1);
+        a.getOkHttpClient().newCall(new Request.Builder().url(NON_PARSABLE_URL).build()).enqueue(new GetMeasurementJsonCallback() {
```
bassosimone: Can you rename `NON_PARSABLE_URL` like `NON_EXISTING_MEASUREMENT_URL` (or even something shorter but meaningful)? The name has now become misleading.

PR #251, opened 2019-08-26
Copy explorer url

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/856

## Proposed Changes

  - Calling the measurements url every time opening this screen (if json file is on disk) and delete it when published. Having in this case the “status” of the publishing of the measurement in this screen
  - Showing different messages if the measurement has been published or not
  - Copies the url to clipboard

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,0 +1,116 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <codeStyleSettings language="XML">
+      <indentOptions>
+        <option name="CONTINUATION_INDENT_SIZE" value="4" />
+      </indentOptions>
+      <arrangement>
+        <rules>
+          <section>
+            <rule>
+              <match>
+                <AND>
+                  <NAME>xmlns:android</NAME>
+                  <XML_ATTRIBUTE />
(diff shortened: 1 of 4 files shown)


.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,116 @@
+<component name="ProjectCodeStyleConfiguration">
```
bassosimone: Is it intended to commit this file?

app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
```diff
@@ -68,7 +71,7 @@
     Toolbar toolbar;
     private Measurement measurement;
     private Snackbar snackbar;
-
+    private Boolean isInExplorer;
```
bassosimone: Please keep an empty line between private variables and methods. I find it more readable.

.idea/codeStyles/Project.xml
lorenzoPrimi: Android studio keep adding it every time i remove it, I guess it's a file needed by the latest update of AS

bassosimone: OK

Issue #252, opened 2019-08-29
Use MKResources to get MMDB and CA


bassosimone:
The relevant implementation is here: https://github.com/measurement-kit/android-libs/blob/master/android-libs/src/main/java/io/ooni/mk/MKResourcesManager.java

We should use this code rather than bundling resources into the repository.

To see how this is done, please check the unit tests, e.g. https://github.com/measurement-kit/android-libs/blob/master/android-libs/src/androidTest/java/io/ooni/mk/androidTest/MKWebConnectivityTest.java

PR #253, opened 2019-08-29
Use MKResources to get MMDB and CA 

lorenzoPrimi:
Fixes https://github.com/ooni/probe-android/issues/252


app/src/main/java/org/openobservatory/ooniprobe/common/OrchestraTask.java
```diff
@@ -36,10 +37,15 @@ public static void sync(Application app) {
 					app.getFilesDir() + "/orchestration_secret.json") ;
 			//TODO ORCHESTRATE
 			//client.setAvailableBandwidth(String value);
-			//what happens when token is nil? should register anyway with empry string
-			client.setCABundlePath(app.getCacheDir() + "/" + Application.CA_BUNDLE);
-			client.setGeoIPCountryPath(app.getCacheDir() + "/" + Application.COUNTRY_MMDB);
-			client.setGeoIPASNPath(app.getCacheDir() + "/" + Application.ASN_MMDB);
+			//what happens when token is nil? should register anyway with empty string
```
bassosimone: What token?

app/src/main/java/org/openobservatory/ooniprobe/test/TestAsyncTask.java
```diff
@@ -54,9 +56,15 @@ protected TestAsyncTask(ACT activity, Result result) {
...
 					MKGeoIPLookupTask geoIPLookup = new MKGeoIPLookupTask();
 					geoIPLookup.setTimeout(act.getResources().getInteger(R.integer.default_timeout));
-					geoIPLookup.setCABundlePath(act.getCacheDir() + "/" + Application.CA_BUNDLE);
-					geoIPLookup.setCountryDBPath(act.getCacheDir() + "/" + Application.COUNTRY_MMDB);
-					geoIPLookup.setASNDBPath(act.getCacheDir() + "/" + Application.ASN_MMDB);
+					boolean okay = MKResourcesManager.maybeUpdateResources(act);
+					if (!okay) {
+						Exception e = new Exception("MKResourcesManager didn't find resources");
+						Crashlytics.logException(e);
+						throw e;
```
bassosimone: Who is going to catch this exception?

app/src/main/java/org/openobservatory/ooniprobe/model/settings/Settings.java
```diff
@@ -81,9 +83,10 @@ public Annotations(Context c) {
 		public Integer all_endpoints;
 
 		public Options(Context c, PreferenceManager pm) {
-			ca_bundle_path = c.getCacheDir() + "/" + Application.CA_BUNDLE;
-			geoip_asn_path = c.getCacheDir() + "/" + Application.ASN_MMDB;
-			geoip_country_path = c.getCacheDir() + "/" + Application.COUNTRY_MMDB;
+			//TODO where to check these are not null?
+			ca_bundle_path = MKResourcesManager.getCABundlePath(c);
+			geoip_country_path = MKResourcesManager.getCountryDBPath(c);
+			geoip_asn_path = MKResourcesManager.getASNDBPath(c);
```
bassosimone: Having seen the code, I believe we have the same behaviour if we just call this method when the app is loaded because that's what the pre-existing code was already doing. I encourage you to do that and then open an issue saying that in some cases we may have issues if the app is open because the user may have cancelled the cache.

I'm leaning towards making the resources loading something inside android-libs in the next release, so all this complexity is not exposed to the application and we try to load them wherever needed from android-libs.

Sounds good?

app/src/main/java/org/openobservatory/ooniprobe/common/OrchestraTask.java
lorenzoPrimi: it's an ancient comment, not to be addressed now

app/src/main/java/org/openobservatory/ooniprobe/test/TestAsyncTask.java
lorenzoPrimi: row 85

app/src/main/java/org/openobservatory/ooniprobe/model/settings/Settings.java
lorenzoPrimi: I added the checks on every run test. check the edits in AbstractTest

app/src/main/java/org/openobservatory/ooniprobe/model/settings/Settings.java
```diff
@@ -81,9 +83,10 @@ public Annotations(Context c) {
 		public Integer all_endpoints;
 
 		public Options(Context c, PreferenceManager pm) {
-			ca_bundle_path = c.getCacheDir() + "/" + Application.CA_BUNDLE;
-			geoip_asn_path = c.getCacheDir() + "/" + Application.ASN_MMDB;
-			geoip_country_path = c.getCacheDir() + "/" + Application.COUNTRY_MMDB;
+			//TODO where to check these are not null?
```
bassosimone: What do you mean with this comment?

lorenzoPrimi: removed. was an old comment

PR #256, opened 2019-09-04
Checking for internet connection before downloading raw data from explorer

lorenzoPrimi:


app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
```diff
@@ -225,6 +226,13 @@ public boolean onOptionsItemSelected(MenuItem item) {
                     json = new GsonBuilder().setPrettyPrinting().disableHtmlEscaping().create().toJson(new JsonParser().parse(json));
                     startActivity(TextActivity.newIntent(this, json));
                 } catch (Exception e) {
+                    if (OrchestraTask.getNetworkType(this).equals(OrchestraTask.NO_INTERNET)) {
```
bassosimone: It feels weird that this functionality is part of the orchestra task. Why is that? Is that cached?

app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
```diff
@@ -225,6 +226,13 @@ public boolean onOptionsItemSelected(MenuItem item) {
...
+                        new MessageDialogFragment.Builder()
+                                .withTitle(getString(R.string.Modal_Error))
+                                .withMessage(getString(R.string.Modal_Error_RawDataNoInternet))
+                                .build().show(getSupportFragmentManager(), null);
+                        return true;
```
bassosimone: I like this pattern you're using here more than the iOS one, wrt structured programming.

app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
lorenzoPrimi: It's where I found it. The orchestra task is where the connection status is stored

app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
lorenzoPrimi: It's the same pattern to launch alert I found around the app

app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
bassosimone: So, I think we need to understand whether this is what we want. Can we maybe check if there is an idiomatic Android way to see if we have connectivity and/or whether we're using that in orchestra? (Also, as I've read on StackOverflow when starting to research this issue, I believe having "internet" does not mean we'll succeed [*]).

[*] However, I guess the purpose of this check is to tell the user they cannot download from the internet when in airplane mode, or in other clearly no-internet cases. Is my understanding correct?

app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
bassosimone: Yeah, I like this because it returns early.

app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
lorenzoPrimi: This is the standard android way. 
Just the function it has been moved inside the OrchestraTask class

lorenzoPrimi: https://github.com/ooni/probe-android/blob/master/app/src/main/java/org/openobservatory/ooniprobe/common/OrchestraTask.java#L71

bassosimone: Please create an issue saying we want the ReachabilityManager back

lorenzoPrimi: https://github.com/ooni/probe-android/issues/257

PR #262, opened 2019-09-25
Fixing max_runtime not being parsed

### LINKED ISSUE #259: Web connectivity:  Test duration is not respected
##  Expected Behavior
Run a Web Connectivity test with a duration >< 120 seconds (for instance 1000 seconds).

## Actual Behavior

Web connectivity runs with the default test duration of 120 seconds.

## Steps to Reproduce the Problem

  1. Tap Web connectivity (Websites) card
  2. Tap on the settings icon
  3. Set the `Test duration` to a non default value

## Specifications

  - Version: 2.1.0
  - Platform: Android 7.1.2
  - Subsystem: measurement-kit 0.10.4

lorenzoPrimi:
Closes #259

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/common/PreferenceManager.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/common/PreferenceManager.java
@@ -45,7 +45,7 @@ public class PreferenceManager {
 
 	public Integer getMaxRuntime() {
 		try {
-			return sp.getInt(r.getString(R.string.max_runtime), 90);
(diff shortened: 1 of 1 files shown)


lorenzoPrimi:
@hellais should this be included in the next 2.2 release?

hellais:
> should this be included in the next 2.2 release?

Yes for sure.

Can we perhaps also add some tests to ensure there is no relapse in this bug? I believe something similar to this was already reported in the past and had we some checks maybe this would have not relapsed (see: https://github.com/ooni/probe/issues/860#issue-451576718).

PR #263, opened 2019-09-30
Refactor Notification in a class

lorenzoPrimi:
I moved the `sendNotificaiton` function in a separate class.
This will help the future implementation of automatic testing in background.
I refactored the `sendNotification` function to be as generic as possible and get a text and an icon, while the `notifyTestEnded` function gets a testSuite and from that will generate text and icon.

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/common/NotificationService.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/common/NotificationService.java
@@ -1,0 +1,46 @@
+package org.openobservatory.ooniprobe.common;
+
+import android.app.Notification;
+import android.app.NotificationChannel;
+import android.app.NotificationManager;
+import android.app.PendingIntent;
+import android.content.Context;
+import android.graphics.Bitmap;
+import android.graphics.Canvas;
+import android.graphics.drawable.Drawable;
+import android.os.Build;
+
+import androidx.core.app.NotificationCompat;
+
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/openobservatory/ooniprobe/common/NotificationService.java
```diff
@@ -0,0 +1,46 @@
...
+    public static void notifyTestEnded(Context c, AbstractSuite testSuite) {
+        sendNotification(c, c.getString(testSuite.getTitle()) + " " + c.getString(R.string.Notification_FinishedRunning), c.getResources().getDrawable(testSuite.getIcon()));
+    }
+
+    public static void sendNotification(Context c, String text, Drawable icon) {
+        NotificationManager notificationManager = (NotificationManager) c.getSystemService(Context.NOTIFICATION_SERVICE);
+        if (notificationManager == null)
+            return;
+        if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O)
+            notificationManager.createNotificationChannel(new NotificationChannel(TEST_RUN, c.getString(R.string.Settings_Notifications_OnTestCompletion), NotificationManager.IMPORTANCE_DEFAULT));
```
bassosimone: I believe you should wrap this line, because it's too long. Because you wrap, also please use `{` and `}` to clearly delimit where the `if` body begins and ends.

Issue #271, opened 2019-11-24
When an upload fails it should not show a toast saying "Upload successful"


hellais:
See the screenshot below:
![Screenshot_20191124-124652](https://user-images.githubusercontent.com/424620/69494206-fa0abb00-0eb8-11ea-8020-e818ac357112.png)

lorenzoPrimi:
Fixed in https://github.com/ooni/probe-android/pull/272

PR #272, opened 2019-11-25
Update ResubmitTask.java

lorenzoPrimi:
Fixes https://github.com/ooni/probe-android/issues/271

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/common/ResubmitTask.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/common/ResubmitTask.java
@@ -122,7 +122,7 @@ public class ResubmitTask<A extends AppCompatActivity> extends NetworkProgressAs
     protected void onPostExecute(Boolean result) {
         super.onPostExecute(result);
         A activity = getActivity();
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/openobservatory/ooniprobe/common/ResubmitTask.java
```diff
@@ -122,7 +122,7 @@ protected Boolean doInBackground(Integer... params) {
     protected void onPostExecute(Boolean result) {
         super.onPostExecute(result);
         A activity = getActivity();
-        if (activity != null) {
+        if (activity != null  && result) {
```
bassosimone: shouldn't this be more readable as `result != null`? Can you also avoid adding two spaces before `&&`?

lorenzoPrimi: I need to check the result is true, not null

bassosimone: My apologies, then :whale:

PR #275, opened 2020-01-03
Allow to enable/disable max_runtime

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/893

Missing the string, at the moment it say 0 as runtime

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/activity/OverviewActivity.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/activity/OverviewActivity.java
@@ -62,6 +62,7 @@ public class OverviewActivity extends AbstractActivity {
 		super.onResume();
 		testSuite.setTestList((AbstractTest[]) null);
 		testSuite.getTestList(getPreferenceManager());
+		//TODO if getRuntime = 0 use another string
 		runtime.setText(getString(R.string.twoParam, testSuite.getDataUsage(), getString(R.string.Dashboard_Card_Seconds, testSuite.getRuntime(getPreferenceManager()).toString())));
 	}
 
--- a/app/src/main/java/org/openobservatory/ooniprobe/activity/OverviewActivity.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/activity/OverviewActivity.java
@@ -12,6 +12,7 @@ import android.widget.ImageView;
 import android.widget.TextView;
 
 import org.openobservatory.ooniprobe.R;
+import org.openobservatory.ooniprobe.common.PreferenceManager;
 import org.openobservatory.ooniprobe.model.database.Result;
 import org.openobservatory.ooniprobe.test.suite.AbstractSuite;
 import org.openobservatory.ooniprobe.test.suite.WebsitesSuite;
@@ -65,7 +66,7 @@ public class OverviewActivity extends AbstractActivity {
 		//TODO convert seconds to minutes and hours when needed
 		//if getRuntime = 0 show one hour
 		Integer runTime = testSuite.getRuntime(getPreferenceManager());
-		if (runTime == 0)
+		if (runTime.equals(PreferenceManager.MAX_RUNTIME_DISABLED))
 			runTime = 3600;
 		runtime.setText(getString(R.string.twoParam, testSuite.getDataUsage(), getString(R.string.Dashboard_Card_Seconds, runTime.toString())));
 	}
--- a/app/src/main/java/org/openobservatory/ooniprobe/activity/OverviewActivity.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/activity/OverviewActivity.java
@@ -12,7 +12,6 @@ import android.widget.ImageView;
 import android.widget.TextView;
 
 import org.openobservatory.ooniprobe.R;
-import org.openobservatory.ooniprobe.common.PreferenceManager;
 import org.openobservatory.ooniprobe.model.database.Result;
 import org.openobservatory.ooniprobe.test.suite.AbstractSuite;
 import org.openobservatory.ooniprobe.test.suite.WebsitesSuite;
--- a/app/src/main/java/org/openobservatory/ooniprobe/activity/OverviewActivity.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/activity/OverviewActivity.java
@@ -62,6 +62,7 @@ public class OverviewActivity extends AbstractActivity {
 		super.onResume();
 		testSuite.setTestList((AbstractTest[]) null);
 		testSuite.getTestList(getPreferenceManager());
+		//TODO convert seconds to minutes and hours when needed
 		//if getRuntime = 0 show one hour
 		Integer runTime = testSuite.getRuntime(getPreferenceManager());
 		if (runTime == 0)
--- a/app/src/main/java/org/openobservatory/ooniprobe/activity/OverviewActivity.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/activity/OverviewActivity.java
@@ -62,8 +62,11 @@ public class OverviewActivity extends AbstractActivity {
 		super.onResume();
 		testSuite.setTestList((AbstractTest[]) null);
 		testSuite.getTestList(getPreferenceManager());
-		//TODO if getRuntime = 0 use another string
-		runtime.setText(getString(R.string.twoParam, testSuite.getDataUsage(), getString(R.string.Dashboard_Card_Seconds, testSuite.getRuntime(getPreferenceManager()).toString())));
+		//if getRuntime = 0 show one hour
+		Integer runTime = testSuite.getRuntime(getPreferenceManager());
+		if (runTime == 0)
+			runTime = 3600;
+		runtime.setText(getString(R.string.twoParam, testSuite.getDataUsage(), getString(R.string.Dashboard_Card_Seconds, runTime.toString())));
 	}
 
 	@Override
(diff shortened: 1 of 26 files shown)


app/src/main/java/org/openobservatory/ooniprobe/fragment/PreferenceFragment.java
```diff
@@ -115,6 +115,8 @@ else if (preference instanceof EditTextPreference) {
                 getFragmentManager().beginTransaction().replace(android.R.id.content, newConcreteInstance(rootKey)).commit();
             }
         } else if (preference instanceof SwitchPreferenceCompat) {
+            if (key.equals(getString(R.string.max_runtime_enabled)))
+                return;
```
bassosimone: I don't fully understand the logic here, can you explain @lorenzoPrimi?

lorenzoPrimi: Because this code written by xan it's to manage the case of "not disable every category" and for some reasons it contrast with this new switch

bassosimone: I still fail to fully understand, sorry 🤷‍♀

bassosimone: I feel like we should not rely on hacks and understand how to better the code.

lorenzoPrimi: It's not an hack, it's the only way to deal with this scenario "at least one enabled" with the `SwitchPreferenceCompat`
This is the algorithm explained: when disabling a category, check every other category, if at least one is enabled return true, else return false and disable this change.
If it's what you want I can move the next two lines in an external function.

app/src/main/java/org/openobservatory/ooniprobe/activity/OverviewActivity.java
```diff
@@ -62,7 +63,12 @@ public static Intent newIntent(Context context, AbstractSuite testSuite) {
 		super.onResume();
 		testSuite.setTestList((AbstractTest[]) null);
 		testSuite.getTestList(getPreferenceManager());
-		runtime.setText(getString(R.string.twoParam, testSuite.getDataUsage(), getString(R.string.Dashboard_Card_Seconds, testSuite.getRuntime(getPreferenceManager()).toString())));
+		//TODO convert seconds to minutes and hours when needed
+		//if getRuntime = 0 show one hour
+		Integer runTime = testSuite.getRuntime(getPreferenceManager());
+		if (runTime.equals(PreferenceManager.MAX_RUNTIME_DISABLED))
+			runTime = 3600;
```
bassosimone: Why setting the runtime to 3600? What impact does it have on the UI?

app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
```diff
@@ -47,7 +48,7 @@
     LottieAnimationView animation;
     private AbstractSuite testSuite;
     private boolean background;
-    private Integer runtime;
+    private static Integer runtime;
```
bassosimone: Why is this now static?

app/src/main/java/org/openobservatory/ooniprobe/item/TestsuiteItem.java
```diff
@@ -43,7 +43,12 @@ public TestsuiteItem(AbstractSuite extra, PreferenceManager pm, View.OnClickList
 		holder.itemView.setOnClickListener(onClickListener);
 		holder.run.setTag(extra);
 		holder.itemView.setTag(extra);
-		holder.runtime.setText(holder.runtime.getContext().getString(R.string.Dashboard_Card_Seconds, extra.getRuntime(pm).toString()));
+		//TODO convert seconds to minutes and hours when needed
+		//if getRuntime = 0 show one hour
+		Integer runTime = extra.getRuntime(pm);
+		if (runTime.equals(PreferenceManager.MAX_RUNTIME_DISABLED))
+			runTime = 3600;
+		holder.runtime.setText(holder.runtime.getContext().getString(R.string.Dashboard_Card_Seconds, runTime.toString()));
```
bassosimone: Maybe we should factor this duplicate code?

app/src/main/java/org/openobservatory/ooniprobe/test/test/WebConnectivity.java
```diff
@@ -35,11 +35,14 @@ public WebConnectivity() {
 	}
 
 	@Override public int getRuntime(PreferenceManager pm) {
-		if (getMax_runtime() != null)
+		if (getMax_runtime() != null  &&
+				!getMax_runtime().equals(PreferenceManager.MAX_RUNTIME_DISABLED))
```
bassosimone: Can we check for the max runtime being >= 0 here?

app/src/main/java/org/openobservatory/ooniprobe/activity/OverviewActivity.java
lorenzoPrimi: Hellais said we should show one hour runtime

app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
lorenzoPrimi: Because otherwise I couldn't use  in a function below

app/src/main/java/org/openobservatory/ooniprobe/test/test/WebConnectivity.java
lorenzoPrimi: Same as iOS PR. You should know the logic of setting runtime by the user more in deep

app/src/main/java/org/openobservatory/ooniprobe/item/TestsuiteItem.java
lorenzoPrimi: factored

app/src/main/java/org/openobservatory/ooniprobe/activity/OverviewActivity.java
```diff
@@ -12,6 +12,7 @@
 import android.widget.TextView;
 
 import org.openobservatory.ooniprobe.R;
+import org.openobservatory.ooniprobe.common.PreferenceManager;
```
bassosimone: Why is this new import needed?

app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
bassosimone: Static variables are very bad, can we avoid using them?

app/src/main/java/org/openobservatory/ooniprobe/test/test/WebConnectivity.java
bassosimone: I actually don't and find it very convoluted

app/src/main/java/org/openobservatory/ooniprobe/activity/OverviewActivity.java
lorenzoPrimi: Removed

app/src/main/java/org/openobservatory/ooniprobe/test/test/WebConnectivity.java
lorenzoPrimi: The user can set values starting from 11 in this field. So I find more readable checking for `MAX_RUNTIME_DISABLED` rather than <= 0

app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
```diff
@@ -22,6 +22,7 @@
 import com.airbnb.lottie.LottieAnimationView;
 
 import org.openobservatory.ooniprobe.R;
+import org.openobservatory.ooniprobe.common.PreferenceManager;
```
bassosimone: Why do you need to import `PreferenceManager` here? As far as I can tell, you're not using it?

app/src/main/java/org/openobservatory/ooniprobe/test/test/WebConnectivity.java
```diff
@@ -34,12 +34,20 @@ public WebConnectivity() {
...
 
+	/*
+ 	if the option max_runtime is already set in the option and is not MAX_RUNTIME_DISABLED let's use it
+ 	else if the input are sets we calculate 5 seconds per input
+ 	at last we check if max_runtime is enabled, in  that case we use the value in the settings
+ 	*/
 	@Override public int getRuntime(PreferenceManager pm) {
-		if (getMax_runtime() != null)
+		if (getMax_runtime() != null  &&
+				getMax_runtime() > PreferenceManager.MAX_RUNTIME_DISABLED)
```
bassosimone: What is the difference between the max runtime you get here and the one we get from the settings? Why do we have two different authorities regarding max runtime in the code base?

app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
lorenzoPrimi: Removed

app/src/main/java/org/openobservatory/ooniprobe/test/test/WebConnectivity.java
lorenzoPrimi: In  the settings there is the user set runtime,  in the test there is the actual runtime. That can be the setting one or 5 * number_urls in case of ooni run or no max_runtime

PR #281, opened 2020-01-15
Updating report ids to fix travis

lorenzoPrimi:


bassosimone:
@lorenzoPrimi the build is still broken!

@hellais how come that some files have "disappeared" from explorer? Is this WAI?

hellais:
> how come that some files have "disappeared" from explorer? Is this WAI?

Can you better explain what you mean by this? What is WAI?

We have recently been rolling out some optimisation to the OONI API queries with @FedericoCeratto and maybe this could have affected this.

It would be very useful if we had a minimal reproducible case to add it to the API integration tests and work on fixing this (if it is indeed a bug).

lorenzoPrimi:
@hellais these two report have disappeared 
```private static final String EXISTING_REPORT_ID = 20190113T202156Z_AS327931_CgoC3KbgM6zKajvIIt1AxxybJ1HbjwwWJjsJnlxy9rpcGY54VH";```
```private static final String EXISTING_REPORT_ID_2 = "20190702T000027Z_AS5413_6FT78sjp5qnESDVWlFlm6bfxxwOEqR08ySAwigTF6C8PFCbMsM";```

As you can see from 
https://api.ooni.io/api/v1/measurements?report_id=20190113T202156Z_AS327931_CgoC3KbgM6zKajvIIt1AxxybJ1HbjwwWJjsJnlxy9rpcGY54VH

I need the ID of two reports we can safely use for the tests

hellais:
Thanks for following up. This is a very important bug to fix. I suggest you don't update your tests to workaround the API bug, but rather wait for us to deploy a fix for this: https://github.com/ooni/backend/issues/278 which is a very high priority bug.

hellais:
The fix has been deployed. I suggest reverting this last commit and checking if travis still fails.

lorenzoPrimi:
I re-ran t ravis on this PR and worked https://github.com/ooni/probe-android/pull/282
Closing this PR without merging

Issue #229, opened 2019-05-29
view log in new tests doesn't work.


xhdix:
`View Log` doesn't work in new tests in the `web connectivity` section. (About two days)

- There is no problem with old tests. 

- I could not get more information.

- Version: Last

bassosimone:
@xhdix does the proper persist with subsequent betas?

xhdix:
@bassosimone Yes.
An interesting point is that this problem occurred without updating this app.

bassosimone:
@lorenzoPrimi could this be related to https://github.com/ooni/probe/issues/865?

lorenzoPrimi:
Could be, but not for every input.

hellais:
@lorenzoPrimi @bassosimone any ideas on this? Are we able to reproduce this issue?

bassosimone:
I can reproduce this on Android, as follows:

1. enable debug logs

2. run performance

3. see that you miss logs for both

The same happens for Web Connectivity.

lorenzoPrimi:
In the new version the "view  log" button is not shown if the log file is empty. 
I installed the play store version and run performance and I don't see the "view log" button.
I will keep investigating to find the problem.

lorenzoPrimi:
This is the reason why web connectivity don't open logs:
2019-12-24 10:20:48.663 8775-8775/org.openobservatory.ooniprobe W/System.err: Caused by: android.os.TransactionTooLargeException: data parcel size 4333312 bytes
2019-12-24 10:20:48.663 8775-8775/org.openobservatory.ooniprobe W/System.err:     at android.os.BinderProxy.transactNative(Native Method)
2019-12-24 10:20:48.663 8775-8775/org.openobservatory.ooniprobe W/System.err:     at android.os.BinderProxy.transact(Binder.java:1127)
2019-12-24 10:20:48.670 8775-8775/org.openobservatory.ooniprobe W/System.err:     at android.app.IActivityManager$Stub$Proxy.startActivity(IActivityManager.java:3754)
2019-12-24 10:20:48.670 8775-8775/org.openobservatory.ooniprobe W/System.err:     at android.app.Instrumentation.execStartActivity(Instrumentation.java:1669)
2019-12-24 10:20:48.670 8775-8775/org.openobservatory.ooniprobe W/System.err: 	... 30 more

lorenzoPrimi:
In case of Performance test, the debug log is written but then when I go to open it it's just not on disk.
[...]

PR #282, opened 2020-01-15
Speeding up textview loading

lorenzoPrimi:
Fixes https://github.com/ooni/probe-android/issues/229 and https://github.com/ooni/probe/issues/865

From
https://stackoverflow.com/questions/12113623/scrollview-is-extremly-slow-on-ics-android-4-0-when-containing-a-long-textview


bassosimone:
@lorenzoPrimi can you not open a PR when travis is broken?

PR #284, opened 2020-01-17
Bouncer failure logic

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/855

### DIFF
--- a/app/src/main/res/layout/item_failed.xml
+++ b/app/src/main/res/layout/item_failed.xml
@@ -1,0 +1,47 @@
+<?xml version="1.0" encoding="utf-8"?>
+<LinearLayout xmlns:android="http://schemas.android.com/apk/res/android"
+	android:layout_width="match_parent"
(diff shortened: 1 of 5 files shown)


app/src/main/res/layout/item_failed.xml
```diff
@@ -0,0 +1,47 @@
+<?xml version="1.0" encoding="utf-8"?>
```
bassosimone: Can you rename this file `failed_iterm.xml` for consistency with the class name?

lorenzoPrimi: I called this item this way to be consistent with the rest of the project. 
Screenshot 2020-01-27 at 10 46 51

And the naming of Android studio: `SettingsFragment` always has a `fragment_settings.xml` file

bassosimone: 👌

PR #287, opened 2020-01-26
Possibility to retest a set of URLs

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/801


app/src/main/java/org/openobservatory/ooniprobe/activity/ResultDetailActivity.java
```diff
@@ -105,6 +110,54 @@ protected void onResume() {
         load();
     }
 
+    @Override
+    public boolean onCreateOptionsMenu(Menu menu) {
+        getMenuInflater().inflate(R.menu.rerun, menu);
```
bassosimone: Why do you need to inflate the menu explicitly here?

app/src/main/java/org/openobservatory/ooniprobe/activity/ResultDetailActivity.java
```diff
@@ -105,6 +110,54 @@ protected void onResume() {
...
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        switch (item.getItemId()) {
+            case R.id.reRun:
+                new ConfirmDialogFragment.Builder()
+                        .withExtra("re_run")
+                        //TODO sub these rows when the string is inserted
+                        //.withMessage(getString(R.string.Modal_ReRun_Websites_Title))
+                        //.withPositiveButton(getString(R.string.Modal_ReRun_Websites_Run))
```
bassosimone: What value does this commented out lines add?

app/src/main/java/org/openobservatory/ooniprobe/activity/ResultDetailActivity.java
```diff
@@ -105,6 +110,54 @@ protected void onResume() {
...
+                        .withMessage(getString(R.string.Modal_DoYouWantToDeleteThisTest))
+                        .withPositiveButton(getString(R.string.Modal_Delete))
+                        .build().show(getSupportFragmentManager(), null);
```
bassosimone: I am missing why we're offering to delete a test when the user clicks on rerun

app/src/main/java/org/openobservatory/ooniprobe/activity/ResultDetailActivity.java
```diff
@@ -105,6 +110,54 @@ protected void onResume() {
...
+                return true;
+            default:
+                return super.onOptionsItemSelected(item);
+        }
+    }
+
+    private void reRunTest() {
+        AbstractSuite testSuite = result.getTestSuite();
+        WebConnectivity test  =  new WebConnectivity();
```
bassosimone: Two spaces before and after `=` here

app/src/main/java/org/openobservatory/ooniprobe/activity/ResultDetailActivity.java
```diff
@@ -105,6 +110,54 @@ protected void onResume() {
...
+                        .withMessage(getString(R.string.Modal_DoYouWantToDeleteThisTest))
+                        .withPositiveButton(getString(R.string.Modal_Delete))
+                        .build().show(getSupportFragmentManager(), null);
+                return true;
+            default:
+                return super.onOptionsItemSelected(item);
+        }
+    }
+
+    private void reRunTest() {
```
bassosimone: Please choose a different name for this method that more clearly convey what it does. (Context: It is my understanding that this method is only used to rerun the websites nettest.)

app/src/main/java/org/openobservatory/ooniprobe/activity/ResultDetailActivity.java
```diff
@@ -105,6 +110,54 @@ protected void onResume() {
...
+            menu.findItem(R.id.reRun).setVisible(false);
+        return super.onPrepareOptionsMenu(menu);
+    }
+
+    @Override
+    public boolean onOptionsItemSelected(MenuItem item) {
+        switch (item.getItemId()) {
+            case R.id.reRun:
+                new ConfirmDialogFragment.Builder()
+                        .withExtra("re_run")
```
bassosimone: The extra you are using here is different from the one that you are using below

app/src/main/java/org/openobservatory/ooniprobe/activity/ResultDetailActivity.java
```diff
@@ -134,8 +187,10 @@ public void onClick(View v) {
 
     @Override
     public void onConfirmation(Serializable extra, int buttonClicked) {
-        if (buttonClicked == DialogInterface.BUTTON_POSITIVE)
+        if (buttonClicked == DialogInterface.BUTTON_POSITIVE && extra.equals("upload"))
             runAsyncTask();
+        else if (buttonClicked == DialogInterface.BUTTON_POSITIVE && extra.equals("rerun"))
```
bassosimone: The extra you are using here is different from the one you are using above

app/src/main/res/menu/rerun.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="utf-8"?>
+<menu xmlns:android="http://schemas.android.com/apk/res/android"
+	xmlns:app="http://schemas.android.com/apk/res-auto">
+<item
+		android:id="@+id/reRun"
+		android:icon="@drawable/rerun"
+		android:title="@string/TestResults_Overview_Tab_Label"
+		app:showAsAction="ifRoom|withText" />
+	<!-- Modal.ReRun.Websites.Run -->
```
bassosimone: What is the purpose of this comment here?

app/src/main/java/org/openobservatory/ooniprobe/activity/ResultDetailActivity.java
lorenzoPrimi: To show the reload icon on the top right

app/src/main/java/org/openobservatory/ooniprobe/activity/ResultDetailActivity.java
lorenzoPrimi: I cant use the `R.string.Modal_ReRun_Websites_Title` asset as it's not present in the code and travis will fail. I will decomment these rows once the  PR is approved and I update the string files

app/src/main/java/org/openobservatory/ooniprobe/activity/ResultDetailActivity.java
lorenzoPrimi: Placeholder strings. Read above answer

app/src/main/java/org/openobservatory/ooniprobe/activity/ResultDetailActivity.java
bassosimone: Ah, so `.inflate(R.menu.rerun, menu)` is adding the menu entry, I see, thanks!

app/src/main/java/org/openobservatory/ooniprobe/activity/ResultDetailActivity.java
bassosimone: Can I see the final PR rather than one where there is magic commented out code?

lorenzoPrimi: Last time we agreed to add strings files only when the PR is approved to have it more readable to you, but if you want I'll add strings here and then let you know

bassosimone: I guess it's not black and white: in this case it's more useful to see a finished PR. Where the strings have mostly a cosmetic value, then it's probably best to defer adding them, as you say.

lorenzoPrimi: Ok will ping you as soon as I add the strings

lorenzoPrimi: Updated

app/src/main/res/menu/rerun.xml
lorenzoPrimi: Updated

PR #288, opened 2020-01-30
New algorithm to calculate runtime

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/931

## Proposed Changes

  - Dropping runtime from `Result` object in  the database
  - Calculating runtime at runtime with this formula `result_runtime = last_row_start_time - first_row_start_time + last_row_runtime`


app/src/main/java/org/openobservatory/ooniprobe/model/database/Result.java
```diff
@@ -94,9 +94,21 @@ public long countOkMeasurements() {
 	public long countAnomalousMeasurements() {
 		return SQLite.selectCountOf().from(Measurement.class).where(Measurement_Table.result_id.eq(id), Measurement_Table.is_rerun.eq(false), Measurement_Table.is_done.eq(true), Measurement_Table.is_failed.eq(false), Measurement_Table.is_anomaly.eq(true)).count();
 	}
+	
+	private Measurement getFirstMeasurement() {
+		return SQLite.select().from(Measurement.class).where(Measurement_Table.result_id.eq(id), Measurement_Table.is_rerun.eq(false)).orderBy(Measurement_Table.start_time, true).querySingle();
+	}
+
+	private Measurement getLastMeasurement() {
+		return SQLite.select().from(Measurement.class).where(Measurement_Table.result_id.eq(id), Measurement_Table.is_rerun.eq(false)).orderBy(Measurement_Table.start_time, false).querySingle();
```
hellais: Can you wrap these lines to a maximum column width to make it more readable?

hellais: I also verified that `querySingle` does the right thing of applying LIMIT 1, and it does according to: https://dbflow.gitbook.io/dbflow/v/develop/usage2/usage/retrieval#synchronous-retrieval

app/src/main/java/org/openobservatory/ooniprobe/model/database/Result.java
```diff
@@ -94,9 +94,21 @@ public long countOkMeasurements() {
...
+	}
 
-	public void addDuration(double value) {
-		this.runtime += value;
+	public double getRuntime(){
+		Measurement first = getFirstMeasurement();
+		Measurement last = getLastMeasurement();
```
hellais: Is there a possibility of these not being defined? What happens if there are no rows for the specified search criteria can do we get an exception or maybe an undefined model?

Can we add some extra checks for this?

app/src/main/java/org/openobservatory/ooniprobe/model/database/Result.java
```diff
@@ -94,9 +94,21 @@ public long countOkMeasurements() {
...
+		long diffInMs = last.start_time.getTime() - first.start_time.getTime();
+		long diffInSec = TimeUnit.MILLISECONDS.toSeconds(diffInMs);
+		return diffInSec + last.runtime;
```
hellais: This algorithm looks good.

app/src/main/java/org/openobservatory/ooniprobe/model/database/Result.java
lorenzoPrimi: You can't get nil because you can't access the test summary screen if there are no measurement in the result. See https://github.com/ooni/probe/issues/855

app/src/main/java/org/openobservatory/ooniprobe/model/database/Result.java
```diff
@@ -110,6 +110,8 @@ private Measurement getLastMeasurement() {
 	public double getRuntime(){
 		Measurement first = getFirstMeasurement();
 		Measurement last = getLastMeasurement();
+		if (first == null || last  == null)
+			return 0;
```
hellais: Same as for iOS, please handle this as an exception, not as a valid runtime value.

PR #290, opened 2020-02-05
Custom websites: sanitize input

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/982

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/activity/CustomWebsiteActivity.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/activity/CustomWebsiteActivity.java
@@ -47,6 +47,7 @@ public class CustomWebsiteActivity extends AbstractActivity implements ConfirmDi
             for (EditText editText : editTexts) {
                 String value = editText.getText().toString();
                 String sanitizedUrl = value.replaceAll("\\r\\n|\\r|\\n", " ");
+                //https://support.microsoft.com/en-us/help/208427/maximum-url-length-is-2-083-characters-in-internet-explorer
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/openobservatory/ooniprobe/activity/CustomWebsiteActivity.java
```diff
@@ -46,8 +46,9 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
             ArrayList<String> urls = new ArrayList<>(editTexts.size());
             for (EditText editText : editTexts) {
                 String value = editText.getText().toString();
-                if (Patterns.WEB_URL.matcher(value).matches() && value.length() < 2084)
-                    urls.add(Url.checkExistingUrl(value).toString());
+                String sanitizedUrl = value.replaceAll("\\r\\n|\\r|\\n", " ");
+                if (Patterns.WEB_URL.matcher(sanitizedUrl).matches() && sanitizedUrl.length() < 2084)
```
bassosimone: Please add the link to the place where you found that 2083 is the maximum URL length for IE :-)

lorenzoPrimi: I have no idea when this 2084 was written. But nice one, I'll add a comment!
```
There is no set minimum or maximum URL length, but some applications do have length restrictions; Internet Explorer imposes a 2084-character limit. Microsoft states that the maximum length of a URL in Internet Explorer is 2,083 characters, with no more than 2,048 characters in the path portion of the URL.
```
https://support.microsoft.com/en-us/help/208427/maximum-url-length-is-2-083-characters-in-internet-explorer

PR #292, opened 2020-02-09
Move settings

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/881


app/src/main/res/values-fr/strings.xml
```diff
@@ -55,8 +55,8 @@
   <string name="Dashboard_InstantMessaging_Card_Description">Testez le blocage des applis de messagerie instantanée</string>
   <string name="Dashboard_InstantMessaging_Overview_Paragraph">Vérifiez si [WhatsApp](https://ooni.io/nettest/whatsapp/), [Facebook Messenger](https://ooni.io/nettest/facebook-messenger/) et [Telegram](https://ooni.io/nettest/telegram/) sont bloqués.\n\nVos résultats seront publiés dans l’[explorateur OONI](https://explorer.ooni.io/world/) et l’[API OONI](https://api.ooni.io/).</string>
   <string name="TestResults_Gbps">Gbps</string>
-  <string name="TestResults_Mbps">Mbps</string>
-  <string name="TestResults_Kbps">Kbps</string>
+  <string name="TestResults_Mbps">Mbit/s</string>
+  <string name="TestResults_Kbps">Kbit/s</string>
```
bassosimone: please, use `kbit/s`

lorenzoPrimi: It's an auto update of strings from transifex, please edit the FR strings there and tell the translator about the why https://www.transifex.com/otf/ooniprobe/translate/#fr/strings-json/158100916

bassosimone: :woman_facepalming:

hellais: Our original source strings also use the same convention: https://github.com/ooni/translations/blob/master/probe-mobile/en/strings.csv#L58.

I think we should be first updating our source strings and then fixing them in the translations. The source of the strings is in this spreadsheet: https://docs.google.com/spreadsheets/d/1GPkr_OyNhXJRTXnseyNbo2P4Du-i_W9kJdqjZG4HVkM/edit#gid=0.

bassosimone: Thanks! Used SI units, i.e., kbit/s, Mbit/s, Gbit/s.

lorenzoPrimi: Let me know if I need to update string on transifex, comment there or do anything.
I will merge this PR as soon as we finish this conversation as this is unrelated

bassosimone: I modified the spreadsheet; don't know what needs to happen next. If you can sync up strings maybe do that here, otherwise later. I am not sad if the correct string lands in master in some time from now. It made me sad to use incorrect string and not fix it.

lorenzoPrimi: I will download the spreadsheet and submit it to transifex, either we correct the strings ourselves one by one for every language or we wait the for translators to do it

PR #300, opened 2020-03-05
Interrupt test

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/1026

## Proposed Changes
- Adding an X button in the running screen
- If the test is interruptable send the signal to the go library
- Tell the AsyncTask to don't execute next tasks


app/src/main/java/org/openobservatory/ooniprobe/test/TestAsyncTask.java
```diff
@@ -36,6 +36,8 @@
 	public static final String URL = "URL";
 	protected final WeakReference<ACT> ref;
 	private final Result result;
+	private AbstractTest currentTest;
+	public boolean interrupt;
```
bassosimone, line 40: Please, consider the following:

1. make this variable private

2. set this variable inside the `interrupt` method

3. make the `interrupt` method `synchronized`

4. add a method like `isInterrupted` or something which is also `synchronized`

I believe otherwise we'll have possible race condtions.

lorenzoPrimi: `interrupt` can only go from false to true, never to false, why it should be `synchronized`?

PR #306, opened 2020-03-31
Abstract away MK and run Telegram using the Go engine

bassosimone:
Related to https://github.com/ooni/probe-engine/issues/376 and https://github.com/ooni/probe/issues/1017. This is divided in nice steps where I am performing each step in an individual commit.

Closes https://github.com/ooni/probe/issues/1017


lorenzoPrimi:
I guess `engine/build` has to be added to the `.gitignore`

PR #310, opened 2020-04-15
Remove Notification from preferences

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/1090


app/src/fdroid/res/xml/preferences_global.xml
```diff
@@ -0,0 +1,310 @@
+<androidx.preference.PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android"
```
bassosimone, line 1: Why are you adding this file now? Was it missing from a previous commit?

lorenzoPrimi: It's being added under the `fdroid` folder. So it is used for the fdroid build

bassosimone: Yes. The question is why now and not before?

PR #311, opened 2020-04-22
engine/build.gradle: pin to org.ooni:oonimkall:2020.04.22-151604

bassosimone:
Part of https://github.com/ooni/probe-engine/issues/504


bassosimone:
Please, @lorenzoPrimi, also do some testing. (I see there are no automated tests here and perhaps we should instead add some once we're Go only?)

PR #312, opened 2020-04-23
Add support for running all tests

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/1093

## Proposed Changes

  - Allow `RunningActivity` to get an array of `AbstractSuite` in input, instead of a single one
  - Cycle the Array and execute every test sequencially. Quit the screen only when last test is executed
  - (re)Load the UX every cycle

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
@@ -1,6 +1,7 @@
 package org.openobservatory.ooniprobe.activity;
 
 import android.content.Intent;
+import android.os.Build;
 import android.os.Bundle;
 import android.view.animation.Animation;
 import android.widget.ProgressBar;
@@ -81,6 +82,9 @@ public class RunningActivity extends AbstractActivity {
     private void testStart(){
         runtime = testSuite.getRuntime(getPreferenceManager());
         getWindow().setBackgroundDrawableResource(testSuite.getColor());
+        if (Build.VERSION.SDK_INT >= 21) {
+            getWindow().setStatusBarColor(testSuite.getColor());
+        }
         animation.setImageAssetsFolder("anim/");
         animation.setAnimation(testSuite.getAnim());
         animation.setRepeatCount(Animation.INFINITE);
--- a/app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
@@ -29,6 +29,9 @@ import org.openobservatory.ooniprobe.test.TestAsyncTask;
 import org.openobservatory.ooniprobe.test.suite.AbstractSuite;
 import org.openobservatory.ooniprobe.common.OrchestraTask;
 
+import java.util.ArrayList;
+import java.util.List;
+
 import butterknife.BindView;
 import butterknife.ButterKnife;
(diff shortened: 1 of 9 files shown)


app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
```diff
@@ -45,38 +37,61 @@
...
-        testSuite = (AbstractSuite) getIntent().getSerializableExtra(TEST);
-        runtime = testSuite.getRuntime(getPreferenceManager());
-        setTheme(testSuite.getThemeDark());
+        setTheme(R.style.Theme_MaterialComponents_NoActionBar_App);
         setContentView(R.layout.activity_running);
         ButterKnife.bind(this);
+        Bundle extra = getIntent().getBundleExtra(TEST);
+        testSuites = (ArrayList<AbstractSuite>) extra.getSerializable(TEST);
+        if (testSuites == null)
+            finish();
```
bassosimone, line 68: Why aren't you returning here? I think it's weird to `finish` and then `runTest`

app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
```diff
@@ -45,38 +37,61 @@
...
+        runtTest();
+    }
+
+    private void runtTest() {
+        //TODO cycle array
```
bassosimone, line 73: Does this mean that we are still run a single test at a time?

app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
```diff
@@ -45,38 +37,61 @@
...
         setContentView(R.layout.activity_running);
         ButterKnife.bind(this);
+        Bundle extra = getIntent().getBundleExtra(TEST);
+        testSuites = (ArrayList<AbstractSuite>) extra.getSerializable(TEST);
+        if (testSuites == null)
+            finish();
+        runtTest();
+    }
+
+    private void runtTest() {
```
bassosimone, line 72: `runtTest` <- is this intended?

app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
lorenzoPrimi: Old todo, it has been done. Removed

app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
lorenzoPrimi: Took me good 30 seconds to understand the problem! Fixed

app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
lorenzoPrimi: finish() terminates the activity. It's a stronger return

bassosimone: How do you prevent the current thread from calling `runTest`? I've briefly read the documentation of `finish` and I could not infer from the documentation that code placed after `finish` _is not_ run

lorenzoPrimi: I didn't know that, thanks

PR #336, opened 2020-06-17
Implement TorTargets class

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/1181


lorenzoPrimi:
This PR won't be merged as with for these reasons: https://github.com/ooni/probe/issues/1181#issuecomment-646002043

Issue #328, opened 2020-05-25
f-droid: make oonimkall available using JCenter (or similar repos)


licaon-kter:
`ERROR: Found unknown maven repo 'https://dl.bintray.com/ooni/android/' at build.gradle`

These libs are not available [anywhere else](https://gitlab.com/fdroid/fdroidserver/blob/master/fdroidserver/scanner.py#L100)? Then maybe we'll need to build them...

bassosimone:
I tried to make them available to Bintray JCenter and the process failed saying the POM file format was not valid. I did not bother with trying again because that repo was working for us.

```XML
<?xml version="1.0" encoding="UTF-8"?>
<project xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns="http://maven.apache.org/POM/4.0.0"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  <modelVersion>4.0.0</modelVersion>
  <groupId>org.ooni</groupId>
  <artifactId>oonimkall</artifactId>
  <version>2020.05.14-133636</version>
  <packaging>aar</packaging>
</project>
```

[This stackoverflow answer](https://stackoverflow.com/questions/52459232/invalid-pom-project-file-when-sending-request-to-include-bintray-package-in-jcen) highlights some of the reasons why JCEnter could reject a POM. So, maybe I can try and add them, and retry to make the package available through JCenter. (In the past, a POM like that was okay, so I believe they may have made their rules a bit more strict.)

Thanks for letting us know.

bassosimone:
Made some progress by making the JCenter like our POM: https://github.com/ooni/probe-engine/pull/667.

bassosimone:
We should be able to fix this within the next two weeks. Now the package we require has also been added to JCenter. We will fix the `build.gradle` in https://github.com/ooni/probe/issues/1191.

bassosimone:
@licaon-kter AFAICT:

1. the gitlab pipeline [is now building probe-android](https://gitlab.com/fdroid/fdroiddata/-/commit/8c4c05d22b98f98c208e03fc6a741d249c839b8b)

2. my understanding is that version 2.5.0 is now supported by f-droid

3. yet, [the latest available version is 2.3.2](https://f-droid.org/en/packages/org.openobservatory.ooniprobe/)

Is there anything else we should do? Thanks!

licaon-kter:
[...]

PR #341, opened 2020-06-25
Implement Countly

lorenzoPrimi:
Implement https://github.com/ooni/probe/issues/1208 and https://github.com/ooni/probe/issues/1104

## Proposed Changes

  - Implemented Countly for push notification
  - Removed orchestra code
  - Removed crashlytics


app/src/full/AndroidManifest.xml
```diff
@@ -12,9 +12,15 @@
...
-			<intent-filter>
+		<meta-data
+			android:name="com.google.firebase.messaging.default_notification_channel_id"
+			android:value="@string/General_AppName" />
+		<meta-data
+			android:name="firebase_analytics_collection_deactivated"
+			android:value="true" />
+		<service android:name=".common.FirebaseMessagingService"
+			android:exported="false">
+		<intent-filter>
```
bassosimone, line 23: What do these changes do?

app/src/full/java/org/openobservatory/ooniprobe/common/FirebaseMessagingService.java
```diff
@@ -5,32 +5,60 @@
...
+			return;
+		}
+
+		Boolean result = CountlyPush.displayMessage(getApplicationContext(), message, R.drawable.notification_icon, null);
+		if (result == null) {
+			Log.i(TAG, "Message wasn't sent from Countly server, so it cannot be handled by Countly SDK");
+		} else if (result) {
+			Log.i(TAG, "Message was handled by Countly SDK");
+		} else {
+			Log.i(TAG, "Message wasn't handled by Countly SDK because API level is too low for Notification support or because currentActivity is null (not enough lifecycle method calls)");
```
bassosimone, line 37: I think alle the above should be `Log.d`

app/src/full/java/org/openobservatory/ooniprobe/common/FirebaseMessagingService.java
```diff
@@ -5,32 +5,60 @@
...
-		if (remoteMessage.getData().size() > 0) {
-			try {
-				//TODO-FUTURE we can use click_action instead of type
-				JSONObject data = new JSONObject(params.toString());
-			} catch (Exception e) {
-				System.out.println("JSONException " + e);
-			}
+		super.onMessageReceived(remoteMessage);
+
+		Log.d("DemoFirebaseService", "got new message: " + remoteMessage.getData());
```
bassosimone, line 20: Please change the string with which we identify to something different from `DemoFirebaseService`?

app/src/full/java/org/openobservatory/ooniprobe/common/FirebaseMessagingService.java
```diff
@@ -5,32 +5,60 @@
...
+
+		// decode message data and extract meaningful information from it: title, body, badge, etc.
+		CountlyPush.Message message = CountlyPush.decodeMessage(remoteMessage.getData());
+
+		if (message != null && message.has("typ")) {
```
bassosimone, line 25: What is the different between `typ` used here and `type` used below for OONI Run?

app/src/full/java/org/openobservatory/ooniprobe/common/FirebaseMessagingService.java
```diff
@@ -5,32 +5,60 @@
...
+		} else {
+			Log.i(TAG, "Message wasn't handled by Countly SDK because API level is too low for Notification support or because currentActivity is null (not enough lifecycle method calls)");
+		}
+	}
+
+	//Handle ooni run JSON. This is not needed for now
+	public void handleJson(CountlyPush.Message message){
+		Intent notificationIntent = null;
+		if (message.has("type") && message.data("type").equals("ooni_run")) {
+			Log.i(TAG, "It's a OONIRun message!");
```
bassosimone, line 45: Maybe use `Log.d` also here?

app/src/full/java/org/openobservatory/ooniprobe/common/FirebaseMessagingService.java
```diff
@@ -5,32 +5,60 @@
...
+			Log.i(TAG, "Message was handled by Countly SDK");
+		} else {
+			Log.i(TAG, "Message wasn't handled by Countly SDK because API level is too low for Notification support or because currentActivity is null (not enough lifecycle method calls)");
+		}
+	}
+
+	//Handle ooni run JSON. This is not needed for now
+	public void handleJson(CountlyPush.Message message){
+		Intent notificationIntent = null;
+		if (message.has("type") && message.data("type").equals("ooni_run")) {
```
bassosimone, line 44: Maybe using constant strings instead of directly `"type"` and `"ooni_run"`?

app/src/full/java/org/openobservatory/ooniprobe/common/FirebaseMessagingService.java
```diff
@@ -5,32 +5,60 @@
...
+			Log.i(TAG, "It's a OONIRun message!");
+			notificationIntent = new Intent(getApplicationContext(), OoniRunActivity.class);
+			notificationIntent.setFlags(Intent.FLAG_ACTIVITY_NEW_TASK);
+			notificationIntent.putExtra("mv",  message.data("mv"));
+			notificationIntent.putExtra("ta",  message.data("ta"));
+			notificationIntent.putExtra("tn",  message.data("tn"));
```
bassosimone, line 50: I assume the intent is basically not delivered here because of the "not needed" comment, right?

app/src/full/java/org/openobservatory/ooniprobe/common/FirebaseMessagingService.java
```diff
@@ -5,32 +5,60 @@
...
 		}
 	}
 
 	@Override public void onNewToken(String token) {
+		super.onNewToken(token);
 		((Application) getApplicationContext()).getPreferenceManager().setToken(token);
+		CountlyPush.onTokenRefresh(token);
+		Log.i(TAG, "CountlyPush onNewToken " + token); }
```
bassosimone, line 58: Also this should be `Log.d`

app/src/full/java/org/openobservatory/ooniprobe/common/FirebaseMessagingService.java
```diff
@@ -5,32 +5,60 @@
...
+
+	@Override
+	public void onDeletedMessages() {
+		super.onDeletedMessages();
```
bassosimone, line 62: isn't it simpler to _not_ override this method?

app/src/main/AndroidManifest.xml
```diff
@@ -4,6 +4,9 @@
 	<uses-permission android:name="android.permission.INTERNET" />
 	<uses-permission android:name="android.permission.RECEIVE_BOOT_COMPLETED" />
 	<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+	<permission android:name="org.openobservatory.ooniprobe.permission.C2D_MESSAGE" android:protectionLevel="signature" />
+	<uses-permission android:name="org.openobservatory.ooniprobe.permission.C2D_MESSAGE" />
```
bassosimone, line 8: Can you explain to me what these two new permissions do? Any extra permission we could maybe remove?

app/src/main/java/org/openobservatory/ooniprobe/activity/MainActivity.java
```diff
@@ -76,7 +93,24 @@ protected void onNewIntent(Intent intent) {
...
+    public void onConfirmation(Serializable extra, int i) {
+        if (extra.equals(MANUAL_UPLOAD_DIALOG))
+            getPreferenceManager().setManualUploadResults(i == DialogInterface.BUTTON_POSITIVE);
+        else if (extra.equals(ANALYTICS_DIALOG))
+            getPreferenceManager().setSendAnalytics(i == DialogInterface.BUTTON_POSITIVE);
+    }
+
+    @Override
+    public void onStart()
+    {
```
bassosimone, line 105: Can you cuddle up this `{` for consistency with what we do mostly everywhere?

app/src/main/java/org/openobservatory/ooniprobe/activity/MainActivity.java
```diff
@@ -76,7 +93,24 @@ protected void onNewIntent(Intent intent) {
...
+        super.onStart();
+        Countly.sharedInstance().onStart(this);
+    }
+
+    @Override
+    public void onStop()
+    {
```
bassosimone, line 112: Can you cuddle up this `{` for consistency with what we do mostly everywhere?

app/src/main/java/org/openobservatory/ooniprobe/common/CountlyManager.java
```diff
@@ -0,0 +1,56 @@
...
+            ly.count.android.sdk.Countly.CountlyFeatureNames.crashes,
+    };
+
+    public static String[] pushFeatures = new String[]{
+            ly.count.android.sdk.Countly.CountlyFeatureNames.push,
+    };
+
+    public static void register(Context ctx, PreferenceManager preferenceManager){
+        CountlyConfig config = new CountlyConfig()
+                .setAppKey("146836f41172f9e3287cab6f2cc347de3f5ddf3b")
```
bassosimone, line 34: Is it fine to commit the key publicly?

app/src/main/java/org/openobservatory/ooniprobe/common/CountlyManager.java
```diff
@@ -0,0 +1,56 @@
...
+            //TODO evaluate scrolls, clicks, forms, attribution
+    };
+
+    public static String[] crashFeatures = new String[]{
+            ly.count.android.sdk.Countly.CountlyFeatureNames.crashes,
+    };
+
+    public static String[] pushFeatures = new String[]{
+            ly.count.android.sdk.Countly.CountlyFeatureNames.push,
+    };
```
bassosimone, line 30: Could all the above `public static String[]` be private instead?

app/src/main/java/org/openobservatory/ooniprobe/common/ExceptionManager.java
```diff
@@ -1,7 +1,9 @@
 package org.openobservatory.ooniprobe.common;
 
+import ly.count.android.sdk.Countly;
```
bassosimone, line 3: Do I understand correctly that count.ly also works without firebase?

app/src/main/java/org/openobservatory/ooniprobe/common/NotificationService.java
```diff
@@ -5,45 +5,71 @@
...
+        b.setContentTitle(title);
+        b.setContentText(message);
         b.setContentIntent(PendingIntent.getActivity(c, 0, MainActivity.newIntent(c, R.id.testResults), PendingIntent.FLAG_UPDATE_CURRENT));
         notificationManager.notify(1, b.build());
     }
+
+    private static void setToken(Application a){
+        if (a.getPreferenceManager().getToken() != null)
+            CountlyPush.onTokenRefresh(a.getPreferenceManager().getToken());
+        System.out.println("CountlyPush " + a.getPreferenceManager().getToken());
```
bassosimone, line 62: Could you use `Log.d` for that?

build.gradle
```diff
@@ -4,12 +4,11 @@ buildscript {
 		jcenter()
 		mavenCentral()
 		maven { url "https://jitpack.io" }
-		maven { url 'https://maven.fabric.io/public' }
+		maven { url  "http://dl.bintray.com/countly/maven" }
```
bassosimone, line 7: This may or may not cause f-droid issues; see https://github.com/ooni/probe-android/issues/328#issue-623974244.

Isn't it possible to fetch count.ly from bintray's JCenter directly (like we did for probe-engine a while ago)?

app/src/full/AndroidManifest.xml
lorenzoPrimi: https://firebase.google.com/docs/reference/android/com/google/firebase/messaging/FirebaseMessagingService

app/src/full/java/org/openobservatory/ooniprobe/common/FirebaseMessagingService.java
lorenzoPrimi: I copied the code from the Countly website https://support.count.ly/hc/en-us/articles/360037754031-Android-SDK#fcm-integration

app/src/full/java/org/openobservatory/ooniprobe/common/FirebaseMessagingService.java
lorenzoPrimi: That's right, I need a tag

app/src/full/java/org/openobservatory/ooniprobe/common/FirebaseMessagingService.java
lorenzoPrimi: Let's resolve the one above and we unify everything

app/src/full/java/org/openobservatory/ooniprobe/common/FirebaseMessagingService.java
lorenzoPrimi: Should I change it anyway?

app/src/full/java/org/openobservatory/ooniprobe/common/FirebaseMessagingService.java
lorenzoPrimi: it's a countly thing https://support.count.ly/hc/en-us/articles/360037754031-Android-SDK#fcm-integration

app/src/main/java/org/openobservatory/ooniprobe/common/CountlyManager.java
lorenzoPrimi: We always committed the Fabric key

app/src/main/java/org/openobservatory/ooniprobe/common/CountlyManager.java
lorenzoPrimi: Sure

app/src/main/java/org/openobservatory/ooniprobe/common/ExceptionManager.java
lorenzoPrimi: the crash logging yes, it has nothing to do with firebase

app/src/main/java/org/openobservatory/ooniprobe/common/NotificationService.java
lorenzoPrimi: I'm gonna remove this line

build.gradle
lorenzoPrimi: Don't think so https://support.count.ly/hc/en-us/articles/360037754031-Android-SDK#adding-the-countly-sdk

bassosimone: Thanks for sharing the documentation! So, it appears to says the following:

> ### Adding via Android Studio
>
> You may use the default JCenter repository to download the SDK package. If it is not included in your project, include it as follows:
>
> ```Groovy
> buildscript {
>    repositories {
>        jcenter()
>    }
> }
> ```
>
> You can also add the Bintray Maven repository:

My reading of the above is that we probably don't need to manually add their Maven repository and it suffices to fetch their SDK through JCenter that IIRC we are already using. Could you perhaps try and remove their explicit Maven repo and check whether we are able to use Countly anyway through JCenter?

app/src/main/java/org/openobservatory/ooniprobe/common/ExceptionManager.java
lorenzoPrimi: I can build the fdroid flavor on my machine.

build.gradle
lorenzoPrimi: edited and works

engine/src/main/java/org/openobservatory/engine/Engine.java
```diff
@@ -29,6 +29,11 @@ public static String getVersionMK() {
         return io.ooni.mk.MKVersion.getVersionMK();
     }
 
+    /** getUUID returns the UUID4 generated for this client  */
```
bassosimone, line 32: I disagree with this documentation and I am not sure now the usage is correct. This function returns a new UUID4 every time you invoke it. The way in which it's documented and the name of the function seem to suggest that such UUID4 is cached somewhere and reused, which is not the case. So,

1. I strongly suggest to rename this function `newUUID4`

2. I suggest we make sure we're using this function correctly

Regarding 2., I will insert a specific comment where needed

app/src/main/java/org/openobservatory/ooniprobe/common/CountlyManager.java
```diff
@@ -0,0 +1,71 @@
...
+
+    private static String[] pushFeatures = new String[]{
+            ly.count.android.sdk.Countly.CountlyFeatureNames.push,
+    };
+
+    public static void register(Context ctx, PreferenceManager preferenceManager){
+        CountlyConfig config = new CountlyConfig()
+                .setAppKey("146836f41172f9e3287cab6f2cc347de3f5ddf3b")
+                .setContext(ctx)
+                .setDeviceId(Engine.getUUID())
```
bassosimone, line 43: I think what you want here is to get the UUID from the configuration and insert a new UUID into the configuration only if it has not already been inserted. The current code uses a new identity every time.

app/src/full/AndroidManifest.xml
lorenzoPrimi: https://stackoverflow.com/questions/39327141/disable-the-analytics-service-for-firebase-cloud-messaging

app/src/full/java/org/openobservatory/ooniprobe/common/FirebaseMessagingService.java
lorenzoPrimi: copied from the example class

app/src/main/AndroidManifest.xml
lorenzoPrimi: from the countly doc

app/src/main/java/org/openobservatory/ooniprobe/common/ExceptionManager.java
lorenzoPrimi: it is ok as I wrote above

app/src/main/AndroidManifest.xml
lorenzoPrimi: "Add extra lines in AndroidManifest.xml"
https://support.count.ly/hc/en-us/articles/360038842151-Android-SDK-19-02-

app/build.gradle
```diff
@@ -19,12 +19,14 @@ android {
 		testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
 		buildConfigField 'String', 'OONI_ORCHESTRATE_BASE_URL', '"https://orchestrate.ooni.io/"'
```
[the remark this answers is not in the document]
lorenzoPrimi: Needed for get URLS test lists

app/src/full/AndroidManifest.xml
```diff
@@ -12,9 +12,15 @@
...
-		<service android:name=".common.FirebaseMessagingService">
-			<intent-filter>
+		<meta-data
+			android:name="com.google.firebase.messaging.default_notification_channel_id"
+			android:value="@string/General_AppName" />
+		<meta-data
+			android:name="firebase_analytics_collection_deactivated"
+			android:value="true" />
+		<service android:name=".common.FirebaseMessagingService"
+			android:exported="false">
```
lorenzoPrimi, line 22: android:exported Whether or not the broadcast receiver can receive messages from sources outside its application — "true" if it can, and "false" if not. If "false", the only messages the broadcast receiver can receive are those sent by components of the same application or applications with the same user ID.

app/src/full/AndroidManifest.xml
```diff
@@ -12,9 +12,15 @@
 		<meta-data
 			android:name="com.google.firebase.messaging.default_notification_color"
 			android:resource="@color/color_base" />
-
-		<service android:name=".common.FirebaseMessagingService">
-			<intent-filter>
+		<meta-data
+			android:name="com.google.firebase.messaging.default_notification_channel_id"
+			android:value="@string/General_AppName" />
```
bassosimone, line 17: This is the name of the notification channel that will be used by Firebase.

app/src/full/java/org/openobservatory/ooniprobe/common/FirebaseMessagingService.java
bassosimone: I suggest see if we can avoid integrating this "typ" thing to implement the functionality we want. If we need it, then I think we should better document what is this and why we need it.

app/src/full/java/org/openobservatory/ooniprobe/common/FirebaseMessagingService.java
```diff
@@ -5,32 +5,60 @@
...
+			Log.i(TAG, "Message wasn't sent from Countly server, so it cannot be handled by Countly SDK");
+		} else if (result) {
+			Log.i(TAG, "Message was handled by Countly SDK");
+		} else {
+			Log.i(TAG, "Message wasn't handled by Countly SDK because API level is too low for Notification support or because currentActivity is null (not enough lifecycle method calls)");
+		}
+	}
+
+	//Handle ooni run JSON. This is not needed for now
+	public void handleJson(CountlyPush.Message message){
```
bassosimone, line 42: We said with @lorenzoPrimi that we can remove this function here

app/src/main/AndroidManifest.xml
bassosimone: This permission allows the cloud to send messages to the device.

app/src/main/java/org/openobservatory/ooniprobe/activity/MainActivity.java
```diff
@@ -54,13 +60,24 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
                 }
             });
             bottomNavigation.setSelectedItemId(getIntent().getIntExtra(RES_ITEM, R.id.dashboard));
-            if (getPreferenceManager().isManualUploadDialog())
```
bassosimone, line 57: Here we should document why the cases below are not mutually exclusive.

app/src/main/java/org/openobservatory/ooniprobe/common/CountlyManager.java
```diff
@@ -0,0 +1,71 @@
...
+
+    public static String[] getConsentsEnabled(PreferenceManager preferenceManager) {
+        List<String> consents = new ArrayList(Arrays.asList(basicFeatures));
+        if (preferenceManager.isSendCrash())
+            consents.addAll(Arrays.asList(crashFeatures));
+        if (preferenceManager.isSendAnalytics())
+            consents.addAll(Arrays.asList(analyticsFeatures));
+        if (preferenceManager.isNotifications())
+            consents.addAll(Arrays.asList(pushFeatures));
+        return consents.toArray((new String[0]));
```
lorenzoPrimi, line 63: https://stackoverflow.com/questions/4042434/converting-arrayliststring-to-string-in-java

The toArray() method without passing any argument returns Object[]. So you have to pass an array as an argument, which will be filled with the data from the list, and returned. You can pass an empty array as well, but you can also pass an array with the desired size.

app/src/full/java/org/openobservatory/ooniprobe/common/FirebaseMessagingService.java
bassosimone: Please, address this comment by either removing the code or explaining why we need it?

app/src/main/java/org/openobservatory/ooniprobe/activity/MainActivity.java
bassosimone: Please document why these two cases can happen together?

PR #343, opened 2020-07-01
Removing Orchestra and its related classes

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/1104


lorenzoPrimi:
No it wasn't. I also removed that code in the iOS branch.

PR #356, opened 2020-09-01
Fix network count on test deletion

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/1240

## Proposed Changes

  - Deleting Network table when deleting all tests
  - Preventing deletion of the network object if it's in use
  - Deleting the network object after the Result is deleted to avoid crash for SQL CONSTRAINTS


bassosimone (approved):
🐳

PR #358, opened 2020-09-04
[WIP] engine: start wiring in probe-engine's geolocate, resubmit

bassosimone:
Tracking issue is https://github.com/ooni/probe-engine/issues/893.

A bunch of issues emerged when working on this and has been documented
at https://github.com/ooni/probe-engine/issues/893#issuecomment-685688006.

Slight changes were also required to ensure the new code was following
the AutoCloseable and try-with-resources pattern.

We should now evaluate with @lorenzoPrimi if this API is good. I originally
wanted to preserve as much as possible the original API, but this was not
immediately easy to do. Let us see how to move forward using this commit as
a prototype of what we could use in the future.

Note: this commit assumes some names in oonimkall that are most likely
poised to change as a consequence of authoring this commit :^).


bassosimone:
Superseded by https://github.com/ooni/probe-android/pull/360

PR #359, opened 2020-09-08
Ask notification permission in the app

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/1210

## Proposed Changes

Show notification dialog :
  - If the user hasn't enabled them
  - The app open count is > 3
  - The dialog has never been showed


app/build.gradle
```diff
@@ -79,6 +79,7 @@ android {
 
 dependencies {
 	implementation 'androidx.appcompat:appcompat:1.1.0'
+    implementation 'androidx.lifecycle:lifecycle-process:2.1.0'
```
bassosimone, line 82: Please, use the same indent of the lines around here

app/src/main/java/org/openobservatory/ooniprobe/common/PreferenceManager.java
```diff
@@ -98,10 +100,25 @@ public boolean isManualUploadDialog() {
...
+	}
+
 	public boolean isNotifications() {
 		return sp.getBoolean(r.getString(R.string.notifications_enabled), false);
 	}
 
+	public void setNotificationsFromDialog(boolean notifications) {
+		//set notification value and increment app open
+		sp.edit()
+				.putLong(IS_NOTIFICATION_DIALOG, getAppOpenCount()+1)
```
bassosimone, line 117: Why are we incrementing the count here? It does not seem necessary when setting the notifications?

bassosimone (commented):
Asking questions!

app/src/main/java/org/openobservatory/ooniprobe/common/PreferenceManager.java
lorenzoPrimi: We do the same thing in iOS, we increment it to don't do it next time the app is opened.

bassosimone: There are many isomorphic way to implement it, I guess.

bassosimone (approved):

PR #360, opened 2020-09-09
engine: remove MK dependency

bassosimone:
Here's another experiment where we're using the API at https://github.com/ooni/probe-engine/issues/893#issuecomment-689031449. The experiment is not complete, but it is possible to clearly see that we are easily and completely able to create wrappers for the Go code such that the Java code only sees interfaces. This is better than the previous experiment because we can really decouple Java code and Go code, which means we can write lots of tests.

Closes https://github.com/ooni/probe-engine/issues/923.


engine/src/main/java/org/openobservatory/engine/PESession.java
```diff
@@ -0,0 +1,39 @@
...
+    public OONIGeolocateTask newGeolocateTask(long timeout) throws OONIException {
+        try {
+            return new PEGeolocateTask(session.newGeolocateTask(timeout));
+        } catch (Exception exc) {
+            throw new OONIException("session.newGeolocateTask failed", exc);
+        }
+    }
+
+    @Override
+    public OONIProbeServicesClient newMakeSubmitterTask(long timeout) throws OONIException {
```
bassosimone, line 27: This should be `newProbeServicesClient`

app/src/main/java/org/openobservatory/ooniprobe/common/ResubmitTask.java
```diff
@@ -39,38 +40,29 @@
...
-            m.is_upload_failed = false;
-            m.save();
-        } else {
-            Log.w(CollectorTask.class.getSimpleName(), results.getLogs());
+        try {
+            input = FileUtils.readFileToString(file, Charset.forName("UTF-8"));
+            try (OONISubmitMeasurementTask task = submitter.newSubmitMeasurementTask(uploadTimeout)){
+                OONISubmitResults results = task.run(input);
+                String output = results.updatedMeasurement;
+                FileUtils.writeStringToFile(file, output, Charset.forName("UTF-8"));
```
bassosimone, line 54: We can simplify the code by not using `output` here

app/src/main/java/org/openobservatory/ooniprobe/common/ResubmitTask.java
```diff
@@ -107,23 +99,27 @@ protected Boolean doInBackground(Integer... params) {
...
-                break;
-            String paramOfParam = activity.getString(R.string.paramOfParam, Integer.toString(i + 1), Integer.toString(measurements.size()));
-            publishProgress(activity.getString(R.string.Modal_ResultsNotUploaded_Uploading, paramOfParam));
-            Measurement m = measurements.get(i);
-            m.result.load();
-            try {
-                if (!perform(activity, m)){
+        try (OONISession session = Engine.newSession(Engine.getDefaultSessionConfig(getActivity(), BuildConfig.SOFTWARE_NAME, BuildConfig.VERSION_NAME, logger));
+             OONIProbeServicesClient client = session.newMakeSubmitterTask(30);
+             OONIReport submitter = client.run();){
```
bassosimone, line 104: The `run` method of `OONIProbeServicesClient` needs to become `openReport`

app/src/main/java/org/openobservatory/ooniprobe/test/TestAsyncTask.java
```diff
@@ -90,6 +89,7 @@ protected TestAsyncTask(ACT activity, Result result) {
 				}
 			} catch (Exception e) {
 				publishProgress(ERR, act.getString(R.string.Modal_Error_CantDownloadURLs));
+				e.printStackTrace();
```
bassosimone, line 92: Shouldn't we pass the exception to the exception manager here?

engine/src/main/java/org/openobservatory/engine/Engine.java
```diff
@@ -2,81 +2,81 @@
...
+        } catch (Exception exc) {
+            throw new OONIException("cannot start OONI Probe Engine task", exc);
         }
-        return new MKExperimentTaskAdapter(io.ooni.mk.MKAsyncTask.start(settings.serialization()));
     }
 
-    /** maybeUpdateResources updates the bundled resources if needed */
-    public static boolean maybeUpdateResources(Context context) {
-        return io.ooni.mk.MKResourcesManager.maybeUpdateResources(context);
+    /** resolveProbeCC creates a new GeoIP lookup task, runs it and return the probeCC. */
```
bassosimone, line 28: `creates` => `returns`

bassosimone (commented):
I suggested small refactoring actions aimed at improving consistency.

engine/build.gradle
```diff
@@ -20,13 +20,16 @@ android {
             proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
         }
     }
+    compileOptions {
+        sourceCompatibility JavaVersion.VERSION_1_8
+        targetCompatibility JavaVersion.VERSION_1_8
+    }
```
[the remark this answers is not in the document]
lorenzoPrimi: This was added by you, don't ask me why :D

engine/src/main/java/org/openobservatory/engine/ArrayLogger.java
```diff
@@ -0,0 +1,29 @@
+package org.openobservatory.engine;
+
+import java.util.ArrayList;
+
+public class ArrayLogger extends AndroidLogger {
```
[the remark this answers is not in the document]
lorenzoPrimi: Actually I don't care about `Log.d` at all and neither the user as it's a debug flag onlyI just thought was a placeholder class you put there to "make it do something" later

engine/src/main/java/org/openobservatory/engine/Engine.java
lorenzoPrimi: read the entire description

engine/build.gradle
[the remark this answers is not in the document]
bassosimone: I guess this was a mistake or an experiment

engine/src/main/java/org/openobservatory/engine/Engine.java
bassosimone: lol - still I think the documentation should not mention implementation details and only say what this function does

bassosimone (commented):
okay, I guess this is good to go now

engine/build.gradle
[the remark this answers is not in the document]
lorenzoPrimi: btw we have the same code in the app module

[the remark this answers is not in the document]
bassosimone: ah, lol, then it's probably fine to keep it!

PR #362, opened 2020-09-10
Add Green icon and change app name on notification

lorenzoPrimi:
Closes https://github.com/ooni/probe/issues/1246
Closes PR (without merging) https://github.com/ooni/probe-android/pull/357

Changes:
- Added green icon
- Changed name of the OoniRunActivity in APP_NAME (that is the app name with the Probe, Dev, Exp suffix)

I managed to achieve this
IMG_20200910_161922


bassosimone (changes requested):
Please, use forest green (`#228B22`) since this green you chose is too bright IMHO

bassosimone (approved):

bassosimone:
@lorenzoPrimi I'm sure we can further improve it down the line. This colour is good enough that my eyes don't bleed when I see it whereas the previous one was really hurting me when looking at it. LGTM.

PR #368, opened 2020-09-28
Fix crash - null check

lorenzoPrimi:
Fixes `java.lang.NullPointerException: Attempt to invoke virtual method 'boolean java.lang.Object.equals(java.lang.Object)' on a null object reference`


bassosimone (approved):
🐳

PR #369, opened 2020-09-30
engine: use ooni/probe-engine@v0.17.0

bassosimone:
Part of https://github.com/ooni/probe-engine/issues/871


engine/src/main/java/org/openobservatory/engine/NullLogger.java
```diff
@@ -0,0 +1,15 @@
+package org.openobservatory.engine;
+
+final class NullLogger implements OONILogger {
+    @Override
+    public void debug(String message) {
```
lorenzoPrimi, line 5: Why not printing the message on the console?

lorenzoPrimi (approved):
Tested and working

PR #373, opened 2020-10-11
Add build stableFdroid CI

mimi89999:


bassosimone (commented):
Thanks for starting this pull request! Two questions:

1. would adding a `connectedStableFdroidDebugAndroidTest` to `emulator.yml` be useful to check whether we're breaking the f-droid build? https://github.com/ooni/probe-android/blob/master/.github/workflows/emulator.yml

2. in which environment does the f-droid build run? Is it a Docker container perhaps? How can we replicate such an environment here (including extra scripts) such that we can really be confident that a future build will successfully complete in f-droid?

Thanks again! 🚀

.github/workflows/build.yml
```diff
@@ -0,0 +1,19 @@
+name: build
+on:
+  push:
```
bassosimone, line 3:
```suggestion
  push:
  pull_request:
```

lorenzoPrimi:
@bassosimone can this be closed or merged?

bassosimone:
Superseded by https://github.com/ooni/probe-android/pull/408#pullrequestreview-578474313

PR #374, opened 2020-10-12
Update build.gradle to fix Fdroid

lorenzoPrimi:
Part of  https://github.com/ooni/probe/issues/1255

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -13,7 +13,7 @@ android {
 		applicationId 'org.openobservatory.ooniprobe'
 		minSdkVersion 19
 		targetSdkVersion 29
-		versionName '2.7.0'
+		versionName '2.7.1'
 		versionCode 65
 		testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
 		buildConfigField 'String', 'OONI_ORCHESTRATE_BASE_URL', '"https://orchestrate.ooni.io/"'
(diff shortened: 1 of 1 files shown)


bassosimone:
@lorenzoPrimi I have changed to original issue's text to avoid closing https://github.com/ooni/probe/issues/1255 for now. I think we also need to address the more fundamental issue, i.e. that we're not checking whether any commit is breaking our f-droid build. I think that https://github.com/ooni/probe-android/pull/373 is a good step forward.

@lorenzoPrimi we should also tag a new version specifically designed to fix the problem for f-droid, otherwise f-droid users will be stuck with a previous version of OONI, which is not optimal. Since today is Sprint planning day, I think we can have a conversation on this matter as part of our planning meetings.

app/build.gradle
```diff
@@ -14,7 +13,7 @@ android {
 		applicationId 'org.openobservatory.ooniprobe'
 		minSdkVersion 19
 		targetSdkVersion 29
-		versionName '2.7.0'
+		versionName '2.7.1'
```
bassosimone, line 16:
```suggestion
		versionName '2.7.0'
```

It seems we don't need to release.

bassosimone (changes requested):
Thanks for fixing the issue @lorenzoPrimi. I think we should not be bumping the version number at this point and then we're good to go. 🌲

bassosimone (approved):
🐳

PR #383, opened 2020-11-10
chore: start using ams-pg-test.ooni.org for testing

bassosimone:
See https://github.com/ooni/backend/issues/461


bassosimone:
This cannot be done very easily. We are using ams-pg.ooni.org for searching for existing measurements. We could possibly point to measurements that only exist on the testing backend. But then maybe we'll wipe the testing backend (for legitimate reasons) therefore breaking these tests. We are also going to reimplement every access to the backend in probe-engine soon, such that we always use the same policies for accessing the backend. Therefore, at this stage it makes sense to say we are not going to pursue the objective of using ams-pg-test.ooni.org in this PR here.

PR #394, opened 2020-11-25
feat(countly): make exceptions more informative

bassosimone:
I've been playing around with countly recently to ensure that the
code is such that exceptions that are trapped by countly are meaningful
enough to us to understand what is going on.

This investigation led me to understand that my original by the book
approach of teaching engine to throw OONIException, as opposed to just
Exception, was probably wrong here.

We're throwing Exception because Go is throwing Exception. Go cannot
do anything much smarter than that here.

By rewrapping the Exception, we're loosing context. We don't see the
cause of error directly in the first line, but further down below such
that we need to click through a specific error to have more info.

What's even more concerning is that countly truncates the stack trace
and the juicy information is dangerously near the bottom. So, it is
a simple refactoring to remove OONIException and it's also simple if
we wanna reintroduce it later. For now let's remove it, since it's
quite clear it may be potentially damaging in the name of purity.

With this diff, my investigation on what could be done to make countly
more actionable is probably complete. The original tracking issue for
this investigation is https://github.com/ooni/probe-engine/issues/1001.

There is also a registered follow-up action that can be performed later
and concerns making the error returned by probe services discovery
even more actionable https://github.com/ooni/probe-engine/issues/1049.

Fixes # <issue>

## Proposed Changes

  -
  -
  -

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/test/TestAsyncTask.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/test/TestAsyncTask.java
@@ -68,6 +68,7 @@ public class TestAsyncTask<ACT extends AbstractActivity> extends AsyncTask<Abstr
 					catch (Exception e) {
 						e.printStackTrace();
 						ExceptionManager.logException(e);
+						// FALLTHROUGH
 					}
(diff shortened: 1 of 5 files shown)


lorenzoPrimi (approved):
I like it

PR #401, opened 2020-12-10
Add connectivity change monitoring

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/1296

The scenario when I'm on wifi and I lose wifi to go on LTE triggers `onLost(Network network)` so it interrupt the test.
On the other hand the scenario when I'm on LTE and I get WiFi
```
2020-12-10 09:33:09.831 16191-16457/org.openobservatory.ooniprobe.dev I/vvv: connected to LTE
2020-12-10 09:33:21.022 16191-16457/org.openobservatory.ooniprobe.dev I/vvv: connected to WIFI
```
doesn't trigger `onLost` but only `onAvailable`, in this case I won't feel comfortable to interrupt the test

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/test/TestAsyncTask.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/test/TestAsyncTask.java
@@ -1,6 +1,9 @@
 package org.openobservatory.ooniprobe.test;
 
+import android.content.Context;
 import android.content.Intent;
+import android.net.ConnectivityManager;
+import android.net.Network;
 import android.os.AsyncTask;
 import android.util.Log;
 
@@ -49,6 +52,8 @@ public class TestAsyncTask extends AsyncTask<Void, String, Void> implements Abst
 	public AbstractTest currentTest;
 	private boolean interrupt;
 	RunTestService service;
+	private ConnectivityManager manager;
+	private ConnectivityManager.NetworkCallback networkCallback;
 
 	public TestAsyncTask(Application app, ArrayList<AbstractSuite> testSuites, RunTestService service) {
 		this.app = app;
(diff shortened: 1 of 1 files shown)


bassosimone:
Marking the PR as ready for review.

bassosimone:
Marking the PR as ready for review.

bassosimone:
@lorenzoPrimi wrote:

> [...] in this case I won't feel comfortable to interrupt the test

maybe we should have a chat about this.

In the meanwhile I'll take a look at the code that has already been proposed here, thank you!

bassosimone (approved):
🐳 

I think we can try and roll out this feature. I am not fully convinced about continuing the test when we're changing network. This actually applies also to the case where we hop from a given WiFi connection to another WiFi connection. As mentioned, we should have a chat about this topic. But this framework in the meanwhile can be added to the app.

lorenzoPrimi:
> I am not fully convinced about continuing the test when we're changing network.

It's not changing network, it's when you are in LTE and you get WiFi, from a WiFi to another Wifi you will get `onLost(Network network) `

bassosimone:
🐳 🐳

lorenzoPrimi:
According to https://developer.android.com/reference/android/net/ConnectivityManager.NetworkCallback

`onAvailable` =  `Called when the framework connects and has declared a new network ready for use.`
and in this case ready doesn't mean in use
while
`onLost` = `Called when a network disconnects or otherwise no longer satisfies this request or callback.`

`
If the callback was registered with requestNetwork() or registerDefaultNetworkCallback(), it will only be invoked against the last network returned by onAvailable() when that network is lost and no other network satisfies the criteria of the request.`

as we use `registerDefaultNetworkCallback` we only monitor the current network when the test starts and we want to be sure that si the network being lost

PR #416, opened 2021-02-14
fix test summary formatting for blocked riseupvpn gateways

cyBerta:
No github issue assigned to the PR

## Proposed Changes

  - currently test summaries are shown incorrectly: instead of the number of blocked gateways or bridged connections the user sees "$1%s blocked" 
  - this PR fixes it:
  
riseupvpnstats

Also the condition under which an anomaly was shown was wrong. The second commit fixes that by using a strict rule that riseupvpn is already considered as blocked if one gateway fails. 

In a subsequent step I would like to implement in probe-cli (probe-engine) 2 test result fields that indicate the overall numbers of tested openvpn and obfs4 gateways. Having that would allow us in the client to consider RiseupVPN blocked less strictly by assuming that if at least one working openvpn port has been reached, openvpn is not blocked and if at least one obfs4 port has been reached obfs4 is not blocked. It will reduce the number of false positives dramtically. I'll create for that proposal a separate issue (https://github.com/ooni/probe/issues/1354).

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/model/jsonresult/TestKeys.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/model/jsonresult/TestKeys.java
@@ -194,10 +194,13 @@ public class TestKeys {
 					blockedConnection++;
 				}
 			}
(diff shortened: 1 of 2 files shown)


lorenzoPrimi (approved):
Great catch, thank you.
Implemented the same in iOS: https://github.com/ooni/probe-ios/pull/424

PR #419, opened 2021-03-04
Adding Wrappers for checkin and test-list

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/1331


app/src/main/java/org/openobservatory/ooniprobe/common/PreferenceManager.java
```diff
@@ -191,6 +191,17 @@ public String getEnabledCategory() {
 		}
 	}
 
+	public ArrayList<String> getEnabledCategoryArr() {
+		ArrayList<String> list = new ArrayList<>(31);
```
bassosimone, line 195: Why `31`? Can't we just leave it empty?

engine/src/main/java/org/openobservatory/engine/OONICheckInConfig.java
```diff
@@ -0,0 +1,73 @@
+package org.openobservatory.engine;
```
bassosimone, line 1: I thought we wanted to remove the `engine` wrappers?

app/src/main/java/org/openobservatory/ooniprobe/common/PreferenceManager.java
lorenzoPrimi: that make sense, I just copied the old code that init the size at 31 made from Ale

lorenzoPrimi: Oh I see:
`
When you create an object of ArrayList in Java without specifying a capacity, it is created with a default capacity which is 10. Since ArrayList is a growable array, it automatically resizes when the size (number of elements in array list) grows beyond a threshold.
`

engine/src/main/java/org/openobservatory/engine/OONICheckInConfig.java
lorenzoPrimi: Not now, I just added more classes to the engine that is another task https://github.com/ooni/probe/issues/1366

bassosimone (approved):
🐳

bassosimone:
(Same here with respect to the tests)

bassosimone:
I see we still have failing tests. I see the failing test is unrelated, so I guess merging is fine. I would be great if we could invest some time into making this test right or into disabling it when running inside a CI environment, so we are not haunted by false positives.

PR #424, opened 2021-04-11
fix app crash when opening MeasurementDetailActivity

cyBerta:
Fixes https://github.com/ooni/probe/issues/1442

## Proposed Changes

  - updates retrofit dependencies
  - logging-interceptor was not needed to update to fix the crash, but was outdated, too.  In order to avoid any backward compatibility issues that cannot be seen on the first glance I updated the logging-interceptor, too


cyBerta:
@lorenzoPrimi another thing: I didn't dived into it much, but the sdk versions differ between the different build.gradle (engine vs. app) files. I'm not sure but it might also cause some trouble.

lorenzoPrimi:
@cyBerta thanks for the PR. I'll look into it, thanks

PR #425, opened 2021-04-11
adaptions for riseupvpn measurement v 0.2.0

cyBerta:
part of https://github.com/ooni/probe/issues/1354

## Proposed Changes

  - OONI Probe Android considers a measurement as anomaly if a all gateways supporting a transport are not working (before: if a single gateway fails)
  - the change is implemented with the intention to reduce false positives if one gateway is not working for whatever reason. In reality RiseupVPN users will fallback to other gateways and might be able to connect though those.


lorenzoPrimi:
@cyBerta thanks for the PR, I'll replicate on iOS as well

lorenzoPrimi:
@cyBerta can you review this PR https://github.com/ooni/probe-ios/tree/riseupvpn_test_0.2 ?
Thank you

lorenzoPrimi:
Didn't see you did iOS as well, thanks for that!

PR #426, opened 2021-04-21
Convert to Unit Tests

sdsantos:
Closes https://github.com/ooni/probe/issues/1445

Some changes in the CI tasks:

- `build` makes sure the code builds on the `stableFull` and `stableFdroid` flavores, on every push
- `test` runs the unit tests on the JVM, on every push
- `emulator` now runs only the instrumented tests, on the same weekly schedule

We also improved a bit the code coverage configuration, and we're now getting a clearer result. UI tests inflate a bit the results, but now we have a stable base to add tests upon:

Screenshot_2021-04-20_at_11 09 55

The full test suite still isn't much faster, since we're now doing 2 separate runs (unit + instrumented tests). But most of future tests will be unit tests, and those are now very quick to build and run, when developing.

---

Additional note:
- Unit tests are able to touch parts of the Android SDK thanks to [Robolectric](http://robolectric.org), but only some unit tests should require it. Those that do, can build upon the `RobolectricAbstractTest` base class.
- Instrumented tests might still fail occasionally on CI, but we'll be able to tackle that once we move to https://github.com/ooni/probe/issues/1447


app/src/test/java/org/openobservatory/ooniprobe/common/ApplicationTest.java
```diff
@@ -0,0 +1,18 @@
...
+import org.openobservatory.ooniprobe.RobolectricAbstractTest;
+import org.robolectric.RobolectricTestRunner;
+
+import static org.junit.Assert.assertEquals;
+
+@SmallTest
+public class ApplicationTest extends RobolectricAbstractTest {
+    @Test
+    public void packageName() {
+        assertEquals("org.openobservatory.ooniprobe.dev", a.getPackageName());
```
lorenzoPrimi, line 16: The package name of the app is `org.openobservatory.ooniprobe`
the .dev is the package name of the dev flavour that is only used internally

lorenzoPrimi: I would always use mainFull for tests as it's the app that lands in the google play store

README.md
```diff
@@ -62,6 +62,26 @@ Instead to build the app to stay compliant to F-Droid use `fdroidRelease`, conta
 ./gradlew assembleFdroidRelease
 ```
 
+## Testing
+
+Run unit tests 
+
+```
+./gradlew testDevFullDebug
```
lorenzoPrimi, line 70: Change in `./gradlew testMainFullDebug`

.github/workflows/test.yml
```diff
@@ -0,0 +1,11 @@
...
+
+on: [push]
+
+jobs:
+  test:
+    runs-on: macos-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+      - run: ./gradlew testDevFullRelease
```
lorenzoPrimi, line 11: - run: ./gradlew testStableFullRelease

sdsantos:
@lorenzoPrimi I thought about what flavour to use, `dev` or `stable`. I picked `dev` just because it's what runs by default when we run an individual test on Android Studio. But I can revert back to `stable`. 

But that `ApplicationTest` is flavour dependent, which isn't great. It will fail when ran in `dev`. So I'll either delete it (it isn't testing anything important), or change it just check if the package name starts with `org.openobservatory.ooniprobe`.

lorenzoPrimi:
I'd use sable for now.
The flavours are:
- `stable` is the flavor we use for the google play release
- `experimental` for long running experimental internal builds
- `dev` for "one shot" internal release with specific features

Furthermore
- `full` is the main app deployed to google play with every feature
- `fdroid` is stripped out of google and any other proprietary SDK

bassosimone:
Thank you!

PR #428, opened 2021-04-29
chore: use ooni/probe-cli v3.10.0-alpha.1

bassosimone:
See https://github.com/ooni/probe/issues/1439


bassosimone:
Let's merge this. It was working for me locally and it seems also tests are green (even though I suppose the only good tests here are the ones in which we use the emulator).

PR #430, opened 2021-05-03
Dependency injection configuration

Filmaluco:
The following PR contains the base files and configuration to install the [Dagger](https://dagger.dev/) compile-time dependency injection framework.

The refactor to include DI in most class's would be huge, since we will be refactoring the [domain logic](https://github.com/ooni/probe/issues/1448) we decided to make those changes incremental for an easier review and safer update.

Small changes can be found already in the _Application.java_ and _PreferenceManager_ class's as well as in the unit test's.

For more information and the official documentation please consult: https://dagger.dev/dev-guide/

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -121,6 +121,10 @@ dependencies {
 	fullImplementation 'ly.count.android:sdk:20.11.1'
 	fullImplementation 'io.sentry:sentry-android:4.0.0'
 
+// Dependency Injection
+	implementation 'com.google.dagger:dagger:2.35'
+	annotationProcessor 'com.google.dagger:dagger-compiler:2.35'
+
 // Testing
 	// Unit Testing
 	testImplementation 'junit:junit:4.13.2'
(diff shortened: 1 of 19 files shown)


Filmaluco:
This PR also closes: https://github.com/ooni/probe/issues/1446

lorenzoPrimi:
@Filmaluco the code looks great but I don't really understand the dependency injector. Can you explain me why is preferable to use it comparing to the current codebase? Thank you

Filmaluco:
@lorenzoPrimi There are various advantages on using a dependency injector framework, but the most fundamentals ones are to more easily inject in test environments and allow and more structured and scalable codebase.

As we move from UITests to the refactoring UseCases and testing them this and other changes will become more obvious a quick example would be the application context, the current codebase tends to pass this argument from method to method.

_example: Function A does not need context but requires it to pass to Function B_

Dagger will allow to more easily inject the context on the Class that needs it (to Function B) and we can simplify both the call on function A and more easily test Class B.

For instance API calls, when testing we don't want to make real calls to an API.  _ApplicationModule_ provides the real api caller while _TestAppModule_ can provide a mock version where we control the return of every method call. 
Removing the need of and having to setup a state with previous instruction to test something and allowing a more convenient cleaner way to provide the test class.

bassosimone (approved):
This is great, thank you!

PR #431, opened 2021-05-12
added: UI Tests

sdsantos:
Closes https://github.com/ooni/probe/issues/1447

Android UI tests, or functional tests, will act as a safety net for future code refactors. We should write at least one test for the main feature of each screen.

---

Most of the chances are only on testing code. But we did some changes on how the App interacts with the Engine. Since we didn't want to run the actual Engine on the UI tests, we needed to be able to swap in a fake Engine. For that we introduced an EngineProvider and an EngineInterface:
- https://github.com/blocoio/probe-android/blob/uiTesting/feedback/app/src/main/java/org/openobservatory/ooniprobe/test/EngineProvider.java
- https://github.com/blocoio/probe-android/blob/uiTesting/feedback/app/src/main/java/org/openobservatory/ooniprobe/test/EngineInterface.java

And calling the Engine from the App would change from `Engine...` to `EngineProvider.get()...`. Let me know what do you think.


bassosimone:
> Since we didn't want to run the actual Engine on the UI tests, we needed to be able to swap in a fake Engine. For that we introduced an EngineProvider and an EngineInterface [...] let me know what do you think

I find this interesting! I was wondering whether having an Engine with interfaces was a good idea. The code generated from Go is mostly classes rather than interfaces. So, it seems there is a use case for Engine written as such.

(It may be that we can refactor the Go code to produce interfaces directly, cutting the complexity a bit more.)

Also, calling `EngineProvider.get()` seems totally fine to me.

lorenzoPrimi (approved):

PR #434, opened 2021-05-13
Bumping minSdkVersion to 24

lorenzoPrimi:


app/src/main/java/org/openobservatory/ooniprobe/test/suite/AbstractSuite.java
```diff
@@ -80,11 +80,7 @@ public int getIcon() {
 	}
 
 	public int getIconGradient() {
-		if(Build.VERSION.SDK_INT <= Build.VERSION_CODES.M){
-			return icon;
-		}else{
-			return icon_24;
-		}
```
bassosimone, line 87: do you also want to remove `icon` here?

lorenzoPrimi: No, we use it elsewhere

PR #438, opened 2021-06-02
Data Tests

sdsantos:
Closes https://github.com/ooni/probe/issues/1449

----

These are the unit tests for the Data related code. That's database, API, preferences and models.

This is also the last task on our Test Coverage project. Some total statistics:
- 43 UI/Instrumented tests
- 201 Unit tests
- Total code coverage: 73%
- Unit tests code coverage: 37%

We focused more on unit tests, since they are faster, cover more critical parts of the logic and are less brittle. But a big part of this project is UI code so, to improve more the coverage percentage, more UI tests will need to be written in the future.

The are we struggled more to test was around async tasks and background services. Those components have different concerns deeply coupled, from persistence, api, UI and background service code. We could not test them completely, special with unit tests, without doing a significant refactor, which was not our goal for this initial project.

Here's the total test coverage chart:

Screenshot 2021-06-02 at 15 42 00

Looking forward to your feedback 😄 

---

_Note: This PR is based on the PRs we submitted previously. We'll solve merge conflicts and clean it up once the other PRs are approved._


engine/src/main/java/org/openobservatory/engine/OONIContext.java
```diff
@@ -10,7 +10,4 @@ public OONIContext(Context ctx) {
         this.ctx = ctx;
     }
 
-    void cancel() {
-        ctx.cancel();
-    }
```
lorenzoPrimi, line 15: @bassosimone we never use this?

app/src/main/java/org/openobservatory/ooniprobe/model/api/CheckIn.java
```diff
@@ -1,18 +0,0 @@
-package org.openobservatory.ooniprobe.model.api;
-
-import org.openobservatory.ooniprobe.model.database.Url;
```
lorenzoPrimi, line 3: I couldn't find this class anywhere, where it has been moved?

app/src/main/java/org/openobservatory/ooniprobe/model/api/UrlList.java
```diff
@@ -1,9 +0,0 @@
-package org.openobservatory.ooniprobe.model.api;
-
-import org.openobservatory.ooniprobe.model.database.Url;
-
```
lorenzoPrimi, line 4: I couldn't find this class anywhere, where it has been moved?

lorenzoPrimi (changes requested):
Left two comments

app/src/main/java/org/openobservatory/ooniprobe/model/api/CheckIn.java
Filmaluco: This class was removed on a cleanup of the project, we did this because having unused code will affect the code coverage, and there's no point in covering this with tests if it's not used... There were quite a few cases of Declaration redundancy.

So I ran Analyze->Inspect Code and in Declaration redundancy I removed all unused and redundant classes, methods, etc.... This was one of them.

app/src/main/java/org/openobservatory/ooniprobe/model/api/UrlList.java
Filmaluco: This class was removed on a cleanup of the project, we did this because having unused code will affect the code coverage, and there's no point in covering this with tests if it's not used... There were quite a few cases of Declaration redundancy.

So I ran Analyze->Inspect Code and in Declaration redundancy I removed all unused and redundant classes, methods, etc.... This was one of them.

app/src/main/java/org/openobservatory/ooniprobe/model/api/CheckIn.java
lorenzoPrimi: Apparently it has been substituted by `org.openobservatory.engine.OONICheckInResults`

app/src/main/java/org/openobservatory/ooniprobe/model/api/UrlList.java
lorenzoPrimi: Same as above

lorenzoPrimi (approved):
You can merge after @bassosimone clarify about the cancel

bassosimone:
@lorenzoPrimi I think it's fine to remove unused code, I therefore merged the PR!

PR #443, opened 2021-07-07
validate URLs

lorenzoPrimi:
Fixes #1706

## Proposed Changes

  - Prevent app to crash when an url is null
  - Validate url in OONI Run


bassosimone (commented):
I'd also try to write a test for this error condition to make sure it does not occur again.

lorenzoPrimi:
@bassosimone Can I merge it and you do it in a second moment?

PR #450, opened 2021-09-28
 RuntimeException after RiseupVPN finishes running

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/1787

Solution from https://stackoverflow.com/questions/58408832/nosuchmethoderror-no-interface-method-getordefaultljava-lang-objectljava-lang

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/common/MapUtility.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/common/MapUtility.java
@@ -1,0 +1,10 @@
+package org.openobservatory.ooniprobe.common;
+
+import java.util.HashMap;
+
+public class MapUtility {
+    //Class replacement for HashMap's function getOrDefault that doesn't work in android < N
+    public static Object getOrDefaultCompat(HashMap map, Object key, Object defaultValue) {
+        return ((map.get(key) != null) ? map.get(key) : defaultValue);
+    }
+}
--- a/app/src/main/java/org/openobservatory/ooniprobe/common/MapUtility.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/common/MapUtility.java
@@ -3,7 +3,7 @@ package org.openobservatory.ooniprobe.common;
 import java.util.HashMap;
 
 public class MapUtility {
-    //Class replacement for HashMap's function getOrDefault that doesn't work in android < N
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/openobservatory/ooniprobe/test/test/RiseupVPN.java
```diff
@@ -27,8 +29,18 @@ public RiseupVPN() {
...
+                    getOrDefault(json.test_keys.transport_status, "obfs4", "ok").equals("blocked");
         measurement.is_anomaly = !json.test_keys.ca_cert_status || json.test_keys.api_failure != null || isTransportBlocked;
     }
+
+    public Object getOrDefault(HashMap map, Object key, Object defaultValue) {
+        //ask your app running more modern API as level 24 (Build.VERSION_CODES.N(ougat))
+        return (android.os.Build.VERSION.SDK_INT >= android.os.Build.VERSION_CODES.N) ?
+                map.getOrDefault(key, defaultValue) :
+        // if not, then need to solve with similar code of original code in next below
+        ((map.get(key) != null) ? map.get(key) : defaultValue);
```
hellais, line 43: I would suggest:
* Using this function independently from the version of Android so we can ensure the code is being tested and run as often as possible
* Moving it into a separate utility related module
* Changing the name to prevent clash with the builtin function (maybe `getOrDefaultCompat`?)

lorenzoPrimi: So you would remove the if and always use the last part of the function?

hellais: Yes that's what I was suggesting.

app/src/main/java/org/openobservatory/ooniprobe/common/MapUtility.java
```diff
@@ -0,0 +1,10 @@
+package org.openobservatory.ooniprobe.common;
+
+import java.util.HashMap;
+
+public class MapUtility {
+    //Class replacement for HashMap's function getOrDefault that doesn't work in android < N
```
hellais, line 6: Can we put the actual version of Android where it doesn't work?

Otherwise looks good to merge.

lorenzoPrimi: I would leave it here so we know when we can remove this class.

PR #461, opened 2021-11-24
fix(RunTestService.java): pending intent should be IMMUTABLE

bassosimone:
* fix(RunTestService.java): pending intent should be IMMUTABLE

The app was crashing because the code did not specify the pending
intent as IMMUTABLE (as far as I understand the stacktrace).

Closes https://github.com/ooni/probe/issues/1897

(This is not my domain of expertise and I think this patch could
and should be better refactored so that we always use a single
factory to create all the intents with the correct flags.)

* fix(RunTestService.java): use factories to avoid repeating ourselves

This diff addresses my previous concern that I fixed the immediate
problem but did not proactively check the rest of the codebase.

Now, `git grep -i pendingintent.newactivity` and
`git grep -i pendingintent.getbroadcast` say that we
have correctly wrapped all the calls.

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/common/service/RunTestService.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/common/service/RunTestService.java
@@ -25,8 +25,6 @@ import org.openobservatory.ooniprobe.test.suite.AbstractSuite;
 
 import java.util.ArrayList;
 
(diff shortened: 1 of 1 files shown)


lorenzoPrimi (approved):
Code looks good.
Works on Android 10

PR #462, opened 2021-11-24
chore: use ooni/probe-cli 3.11.0-beta.2 

bassosimone:
See https://github.com/ooni/probe/issues/1879


bassosimone:
@lorenzoPrimi please, take a look!

PR #463, opened 2021-11-25
don't start automated-testing if the battery level is less than 20%

lorenzoPrimi:
Fixes https://github.com/ooni/probe/issues/1896

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/common/ReachabilityManager.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/common/ReachabilityManager.java
@@ -61,4 +61,14 @@ public class ReachabilityManager {
                 status == BatteryManager.BATTERY_STATUS_FULL;
         return isCharging;
     }
+
+    /* From https://developer.android.com/training/monitoring-device-state/battery-monitoring#DetermineChargeState */
+    public static float getChargingLevel(Context context) {
+        IntentFilter ifilter = new IntentFilter(Intent.ACTION_BATTERY_CHANGED);
+        Intent batteryStatus = context.registerReceiver(null, ifilter);
+        int level = batteryStatus.getIntExtra(BatteryManager.EXTRA_LEVEL, -1);
+        int scale = batteryStatus.getIntExtra(BatteryManager.EXTRA_SCALE, -1);
+        float batteryPct = level * 100 / (float)scale;
+        return batteryPct;
+    }
 }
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/openobservatory/ooniprobe/domain/GenerateAutoRunServiceSuite.java
```diff
@@ -77,7 +78,8 @@ public boolean shouldStart(Boolean isWifi, Boolean isCharging, Boolean isVPNInUs
             return false;
         if (pm.testChargingOnly() && !isCharging)
             return false;
-
+        if(ReachabilityManager.getChargingLevel(app) < 20)
```
hellais, line 81: Should we perhaps still run the test if the phone is charging irrespective of the battery level?

lorenzoPrimi: Fixed

app/src/main/java/org/openobservatory/ooniprobe/common/ReachabilityManager.java
```diff
@@ -61,4 +61,14 @@ public static Boolean isCharging(Context context) {
                 status == BatteryManager.BATTERY_STATUS_FULL;
         return isCharging;
     }
+
+    /* From https://developer.android.com/training/monitoring-device-state/battery-monitoring#DetermineChargeState */
+    public static float getChargingLevel(Context context) {
+        IntentFilter ifilter = new IntentFilter(Intent.ACTION_BATTERY_CHANGED);
+        Intent batteryStatus = context.registerReceiver(null, ifilter);
+        int level = batteryStatus.getIntExtra(BatteryManager.EXTRA_LEVEL, -1);
+        int scale = batteryStatus.getIntExtra(BatteryManager.EXTRA_SCALE, -1);
```
bassosimone, line 70: If both `getInExtra` return `-1` the `batteryPct` is `-1 * 100 / (float) -1` which is equal to `100`. I think we should dd an explicit case here `if (level < 0 || scale < 0) { return 0; }`.

lorenzoPrimi: I used the code from google https://developer.android.com/training/monitoring-device-state/battery-monitoring#DetermineChargeState

bassosimone: okay, then, we're going to assume this is fine since it comes from a blog post in developer.android.com

PR #471, opened 2021-12-17
doc(android): add TODOs pointing to GH issues

bassosimone:
See https://github.com/ooni/probe/issues/1923 and
https://github.com/ooni/probe/issues/1922.

PR #472, opened 2022-01-10
Adds Link to OONI Blog & OONI Reports from About Page

aanorbel:
see ooni/probe#1850

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/activity/InfoActivity.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/activity/InfoActivity.java
@@ -37,10 +37,10 @@ public class InfoActivity extends AbstractActivity {
 	}
 
 	@OnClick(R.id.learnMore) void onLearnMoreClick() {
-		startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse("https://ooni.io/")));
+		startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse("https://ooni.org/")));
 	}
 
 	@OnClick(R.id.dataPolicy) void onDataPolicyClick() {
-		startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse("https://ooni.io/about/data-policy/")));
+		startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse("https://ooni.org/about/data-policy/")));
 	}
 }
--- a/app/src/main/java/org/openobservatory/ooniprobe/activity/InfoActivity.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/activity/InfoActivity.java
@@ -28,6 +28,14 @@ public class InfoActivity extends AbstractActivity {
 		version.setText(getString(R.string.version, BuildConfig.SOFTWARE_NAME, BuildConfig.VERSION_NAME));
 	}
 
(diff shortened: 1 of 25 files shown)


aanorbel:
# Preview on device 

| Light Mode  | Dark Mode|
| ------------- | ------------- |
| photo_2022-01-11 13 42 53 | photo_2022-01-11 13 42 58|

agrabeli (commented):
Great work, thank you @aanorbel! 

I'd suggest @hellais reviews as well, as the CI checks appear to be failing.

aanorbel:
CI failure seems to be a recurring issue. same as in ooni/probe-android#471. An issue was created(ooni/probe#1924) by @bassosimone  to investigate and resolve the problem.

bassosimone (approved):
🐳 thank you!

hellais:
Thanks for putting this together. I am just seeing it now.

I think we make the links to the blog and reports, secondary action buttons similar to how the "data policy" link/button is.

That's because it's considered an anti-pattern to have more than one primary action on a certain screen.

aanorbel:
> Thanks for putting this together. I am just seeing it now.
> 
> I think we make the links to the blog and reports, secondary action buttons similar to how the "data policy" link/button is.
> 
> That's because it's considered an anti-pattern to have more than one primary action on a certain screen.

This is quite accurate i think the option just s. I will make the necessary  update.

app/src/main/java/org/openobservatory/ooniprobe/activity/InfoActivity.java
```diff
@@ -28,6 +28,14 @@
...
+	@OnClick(R.id.blog) void onBlogClick() {
+		startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse("https://ooni.org/blog/")));
+	}
+
+	@OnClick(R.id.reports) void onReportsClick() {
+		startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse("https://ooni.org/reports/")));
+	}
+
 	@OnClick(R.id.learnMore) void onLearnMoreClick() {
 		startActivity(new Intent(Intent.ACTION_VIEW, Uri.parse("https://ooni.io/")));
```
hellais, line 40: While you are at it, can you also change the link here to be `ooni.org`?

aanorbel: Just didn't know how to handle that. Will the update.

aanorbel:
> Thanks for putting this together. I am just seeing it now.
> 
> I think we make the links to the blog and reports, secondary action buttons similar to how the "data policy" link/button is.
> 
> That's because it's considered an anti-pattern to have more than one primary action on a certain screen.


Screenshot_20220114_195915

hellais (changes requested):
Can we have the links below the "learn more" button be evenly spaced? 

Currently there is much more spacing between "Blog" and "Reports" compared to "Reports" and "OONI Data Policy". I suggest using the value of spacing that currently exists between "Reports" and "OONI Data Policy" for the other elements as well.

aanorbel:
<details>
<summary>ScreenShot of page</summary>

</p>
</details>

hellais (approved):
LGTM 🐳 

The CI, though, is failing, so we should document that we need to look into fixing that soon too.

PR #473, opened 2022-01-19
Modified run test command to start minimised

aanorbel:
Fixes  ooni/probe#1976 , ooni/probe#1981 & ooni/probe#1998

## Proposed Changes

- Starts `RunTestService`.
- Triggers `ProgressFragment` to display measurement status.
- Add running test indicator missing in measurement list views.
- Change the minimise test icon from X to downward pointing caret


app/src/main/java/org/openobservatory/ooniprobe/fragment/DashboardFragment.java
```diff
@@ -87,9 +88,12 @@ private void setLastTest() {
...
 
 	public void runAll(){
-		Intent intent = RunningActivity.newIntent((AbstractActivity) getActivity(), testSuites);
-		if (intent != null)
-			ActivityCompat.startActivity(getActivity(), intent, null);
+		RunningActivity.runAsForegroundService((AbstractActivity) getActivity(), testSuites);
+		try {
+			((AbstractActivity)getActivity()).bindTestService();
+		} catch (Exception e) {
+			e.printStackTrace();
```
hellais, line 95: Should we do something about exceptions in this case? Perhaps instead of letting them fall-through we should log them to sentry.

hellais:
Did some light testing of this PR inside of the emulator and it works as expected. The only issue I noticed is that the running test indicator is not present when you navigate to more nested views (see: https://github.com/ooni/probe/issues/1981), but I believe this was already a bug so we can address it separately.

I left a minor comment on something to improve WRT to error handling. The rest of the code is a bit beyond my knowledge of android development, but it seems like some opportunities for refactoring some bits into utility methods have been taken, which is a good thing.

For me, once the issue with the error handling is addressed, this is good to merge!

aanorbel:
Added running test indication to measurement list views
| .| .|
| ------------- | ------------- |
| Screenshot_20220128_055803| Screenshot_20220128_055741|

aanorbel:
New icon to minimise running tests.

Screenshot_20220131_155954

hellais:
Can you put it in the left hand corner instead of the right? This seems to be the location of it in some apps I have seen use this pattern, see: https://github.com/ooni/probe/issues/1998#issuecomment-1026016576.

aanorbel:
# Updated minimise icon position

Screenshot_20220131_200905

hellais (approved):
This looks great! Proceed with merging when you see fit! 🐳

PR #476, opened 2022-01-24
Allow test on emulator to fail gracefully on CI

aanorbel:
Allows ooni/probe#1924 to fail gracefully

## Proposed Changes

  - Increase emulator RAM size
  - Allow emulator test to fail [gracefully](https://github.com/ooni/probe-android/runs/5314721868?check_suite_focus=true#step:4:472).


hellais (approved):
LGTM

PR #483, opened 2022-02-08
add log statement to clarify MKException

aanorbel:
Clarifies ooni/probe#1328

## Proposed Changes

  - Add logging statements to clarify the `MKException`

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/test/TestAsyncTask.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/test/TestAsyncTask.java
@@ -149,10 +149,19 @@ public class TestAsyncTask extends AsyncTask<Void, String, Void> implements Abst
             OONISession session = EngineProvider.get().newSession(EngineProvider.get().getDefaultSessionConfig(
                     app, BuildConfig.SOFTWARE_NAME, BuildConfig.VERSION_NAME, new LoggerArray(), proxy));
             OONIContext ooniContext = session.newContextWithTimeout(30);
+
+            ThirdPartyServices.addLogExtra("ooniContext", app.getGson().toJson(ooniContext));
+
             session.maybeUpdateResources(ooniContext);
             OONIURLListConfig config = new OONIURLListConfig();
             config.setCategories(app.getPreferenceManager().getEnabledCategoryArr().toArray(new String[0]));
+
+            ThirdPartyServices.addLogExtra("config", app.getGson().toJson(config));
+
             OONIURLListResult results = session.fetchURLList(ooniContext, config);
+
+            ThirdPartyServices.addLogExtra("results", app.getGson().toJson(results));
+
             if (results.getUrls().size() == 0) {
                 publishProgress(ERR, app.getString(R.string.Modal_Error_CantDownloadURLs));
                 ThirdPartyServices.logException(new MKException(results));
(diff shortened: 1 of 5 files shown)


bassosimone (approved):
🐳 🙏

app/src/main/java/org/openobservatory/ooniprobe/test/TestAsyncTask.java
```diff
@@ -149,10 +149,19 @@ private void downloadURLs() {
...
+
             session.maybeUpdateResources(ooniContext);
             OONIURLListConfig config = new OONIURLListConfig();
             config.setCategories(app.getPreferenceManager().getEnabledCategoryArr().toArray(new String[0]));
+
+            ThirdPartyServices.addLogExtra("config", app.getGson().toJson(config));
+
             OONIURLListResult results = session.fetchURLList(ooniContext, config);
+
+            ThirdPartyServices.addLogExtra("results", app.getGson().toJson(results));
```
hellais, line 163: Is it correct to assume that this will be sent to sentry only in the event of an exception, but in all cases it will not?

hellais (approved):
This looks good to me. I think logging the values returned by the fetchURLList call before and after passing it to the test runner should give us a bit more insight into what is going on.

In looking at this code I also realised that we have explicit checks to ensure that the `results` list is not empty right after we download it. Since the exception is being triggered in the `run` method of the test runner, I suspect that the input lists is getting cleared or lost somewhere between:
```
OONIURLListResult results = session.fetchURLList(ooniContext, config);
```
and
```
	@Override public void run(Context c, PreferenceManager pm, Gson gson, Result result, int index, TestCallback testCallback) {
		Settings settings = new Settings(c, pm);
		run(c, pm, gson, settings, result, index, testCallback);
	}
```

Perhaps we can add an additional log call inside of `app/src/main/java/org/openobservatory/ooniprobe/test/test/WebConnectivity.java` prior to the run so that we can test this hypothesis too?

If it turns out that `results` has URLs in it before, but then it's empty before calling `run`, we can at least isolate the problem to somewhere in the test running code.

app/src/main/java/org/openobservatory/ooniprobe/test/TestAsyncTask.java
```diff
@@ -133,6 +133,10 @@ private void runTest(AbstractTest... tests) {
                 }
                 if (!interrupt) {
                     Log.d(TAG, "run next suite: " + currentSuite.getName() + " test:" + currentTest.getName());
+
+                    ThirdPartyServices.addLogExtra("name", currentTest.getName());
+                    ThirdPartyServices.addLogExtra("_result", app.getGson().toJson(result));
+
```
hellais, line 139: I'm not sure this is the right place and thing we should be logging. What we are most interested in recording is the `input` that are being passed to the test, which it's my understanding that they are part of the `settings` object that is somehow created via the `context` and `preferencesmanager`.

I would suggest instead to log the `settings` object over here: https://github.com/ooni/probe-android/blob/555917c1deb61c6ad98168fb1a5669b54f000eff/app/src/main/java/org/openobservatory/ooniprobe/test/test/WebConnectivity.java#L24

aanorbel: I thought of doing the log there but i realized that if the offender is not the `WebConnectivity` test, the only other way to capture the test name is to log the name and input here since the `WebConnectivity` and other tests will eventually get called based on their implementations of `AbstractTest.run`

hellais: That's a good point, however we do know that the offender is `WebConnectivity` because:
1. The logs all report the WebConnectivity in the backtrace
2. The only test which takes inputs is currently WebConnectivity

aanorbel: OK i get your point there. 

I went for the generic option that will both clarify the current error with `WebConnectivity` and also any future errors for any other test.

hellais: Excellent, thanks!

hellais:
LGTM 🐳

PR #486, opened 2022-02-11
Improve performance of test startup time

aanorbel:
Fixes ooni/probe#2012

## Proposed Changes

  - Change strategy to save/update urls to bulk save/update to reduce number of queries.
  - ...

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/model/database/Url.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/model/database/Url.java
@@ -23,9 +23,11 @@ import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 
 @Table(database = AppDatabase.class)
 public class Url extends BaseModel implements Serializable {
@@ -110,11 +112,13 @@ public class Url extends BaseModel implements Serializable {
 		}
 
 		if (existingUrls.size() != resultUrlsMap.size()) {
+			Set<String> existingUrlSet = new HashSet<>(Lists.newArrayList(Iterables.transform(existingUrls, input -> input.url)));
+			Set<String> newUrlSet = new HashSet<>(resultUrlsMap.keySet());
+			newUrlSet.removeAll(existingUrlSet);
 
-			List<String> existingUrlStrings = Lists.newArrayList(Iterables.transform(existingUrls, input -> input.url));
-			Map<String, Url> newResultUrls = Maps.filterEntries(resultUrlsMap, input -> !existingUrlStrings.contains(input.getKey()));
+			List<Url> urlsToSave = Lists.transform(Lists.newArrayList(newUrlSet), resultUrlsMap::get);
 
-			Url.saveAll(newResultUrls.values());
+			Url.saveAll(urlsToSave);
 
 		}
 
--- a/app/src/main/java/org/openobservatory/ooniprobe/model/database/Url.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/model/database/Url.java
@@ -48,11 +48,11 @@ public class Url extends BaseModel implements Serializable {
 		return SQLite.select().from(Url.class).where(Url_Table.url.eq(input)).querySingle();
 	}
 
-	public static List<Url> getExistingUrls(Collection<String> urls) {
+	private static List<Url> getExistingUrls(Collection<String> urls) {
 		return SQLite.select().from(Url.class).where(Url_Table.url.in(urls)).queryList();
 	}
 
-	public static void saveAll(Collection<Url> urls) {
+	private static void saveAll(Collection<Url> urls) {
 		final DatabaseDefinition database = FlowManager.getDatabase(AppDatabase.class);
 		database.executeTransaction(databaseWrapper -> Objects.requireNonNull(database.getModelAdapterForTable(Url.class)).saveAll(urls));
 	}
@@ -75,6 +75,10 @@ public class Url extends BaseModel implements Serializable {
 		return url;
 	}
 
+	// This function saves or updates input urls into the database and returns the
+	// list of strings used internally to perform this operation, where each string
+	// is the `.url` value of each URL. We return this list of strings because they are
+	// actually useful to continue processing the input URLs.
 	public static List<String> saveOrUpdate(List<Url> urls) {
 
 		Map<String, Url> resultUrlsMap = Maps.uniqueIndex(
@@ -86,14 +90,13 @@ public class Url extends BaseModel implements Serializable {
 
 		List<Url> existingUrlsToUpdate = Lists.newArrayList(Iterables.filter(existingUrls, input -> {
 			Url incomingChanges = resultUrlsMap.get(input.url);
-			if (incomingChanges != null) {
-				return ((!input.category_code.equals(incomingChanges.category_code)
-						&& !incomingChanges.category_code.equals("MISC"))
-						|| (!input.country_code.equals(incomingChanges.country_code)
-						&& !incomingChanges.country_code.equals("XX")));
-			} else {
+			if (incomingChanges == null) {
 				return false;
 			}
+			return ((!input.category_code.equals(incomingChanges.category_code)
+					&& !incomingChanges.category_code.equals("MISC"))
+					|| (!input.country_code.equals(incomingChanges.country_code)
+					&& !incomingChanges.country_code.equals("XX")));
 		}));
 
 		if (!existingUrlsToUpdate.isEmpty()) {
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/openobservatory/ooniprobe/model/database/Url.java
```diff
@@ -38,6 +48,15 @@ public static Url getUrl(String input) {
 		return SQLite.select().from(Url.class).where(Url_Table.url.eq(input)).querySingle();
 	}
 
+	public static List<Url> getExistingUrls(Collection<String> urls) {
```
bassosimone, line 51:
```suggestion
	private static List<Url> getExistingUrls(Collection<String> urls) {
```

This function does not seem to be used publicly

app/src/main/java/org/openobservatory/ooniprobe/model/database/Url.java
```diff
@@ -38,6 +48,15 @@ public static Url getUrl(String input) {
...
+		return SQLite.select().from(Url.class).where(Url_Table.url.in(urls)).queryList();
+	}
+
+	public static void saveAll(Collection<Url> urls) {
```
bassosimone, line 55:
```suggestion
	private static void saveAll(Collection<Url> urls) {
```

This function does not seem to be used publicly

app/src/main/java/org/openobservatory/ooniprobe/model/database/Url.java
```diff
@@ -56,6 +75,43 @@ public static Url checkExistingUrl(String input, String categoryCode, String cou
...
+		List<Url> existingUrlsToUpdate = Lists.newArrayList(Iterables.filter(existingUrls, input -> {
+			Url incomingChanges = resultUrlsMap.get(input.url);
+			if (incomingChanges != null) {
+				return ((!input.category_code.equals(incomingChanges.category_code)
+						&& !incomingChanges.category_code.equals("MISC"))
+						|| (!input.country_code.equals(incomingChanges.country_code)
+						&& !incomingChanges.country_code.equals("XX")));
+			} else {
+				return false;
+			}
```
bassosimone, line 96:
```suggestion
			if (incomingChanges == null) {
				return false;
			}
			return ((!input.category_code.equals(incomingChanges.category_code)
					&& !incomingChanges.category_code.equals("MISC"))
					|| (!input.country_code.equals(incomingChanges.country_code)
					&& !incomingChanges.country_code.equals("XX")));
```

We don't have a coding style manual that specifies that, however in the past both me and @hellais have expressed preference for not writing an `if .. else` chain where both arms return. The argument is that we'd like to have a single `if` whose arm expresses the less-interesting, often-but-not-in-this-case error related situation and the code falling through expressing the more interesting case.

app/src/main/java/org/openobservatory/ooniprobe/model/database/Url.java
```diff
@@ -56,6 +75,43 @@ public static Url checkExistingUrl(String input, String categoryCode, String cou
 		return url;
 	}
 
+	public static List<String> saveOrUpdate(List<Url> urls) {
```
bassosimone, line 78: This function is named `saveOrUpdate` _but_ it also returns a list of strings from the list of URLs. Therefore I suggest to either document this extra operation that we also perform (which is probably done here to avoid recomputing twice the same value) or to improve the function naming to clarify what it does. If I am given a function named `saveOrUpdate` without extra documentation, I'd be a bit surprised that it _also_ returns something because I'd expect it to just save or update.

If you choose to go down the commenting road, here's an example of what I mean (feel free to write a better Javadoc string if you find the one that I have written is not so good, I meant to write it as an example):

```suggestion
	// This function saves or updates input urls into the database and returns the
	// list of strings used internally to perform this operation, where each string
	// is the `.url` value of each URL. We return this list of strings because they are
	// actually useful to continue processing the input URLs.
	public static List<String> saveOrUpdate(List<Url> urls) {
```

bassosimone (changes requested):
Thanks for writing this pull request! I am particularly happy to see you adopted a functional-like style with modifiers for preparing it! 🔝🔝🔝 

(I am particularly fond of the functional style because it expresses the intent as opposed to writing down in a more assembly-like way the actual steps to be performed.)

I did not spend much time in thinking how bad this transformation would be in terms of memory (and I am not sure I would be able to do that properly). It seems this algorithm is linear multiplied to a constant in the number of URLs. AFAICT, the number of URLs could be around 1-2 thousands. Do you think we should be concerned about that? By heart, I do not think so, but I felt it was still worth asking as part of the review.

Below I have provided some recommendations regarding how to make the code more clear or more in line with implicit guidelines we previously discussed with @hellais regarding writing code but that AFAIK we never wrote down properly. (It would be cool eventually to write them somewhere, perhaps in `CONTRIBUTING.m`.)

app/src/main/java/org/openobservatory/ooniprobe/model/database/Url.java
```diff
@@ -56,6 +75,52 @@ public static Url checkExistingUrl(String input, String categoryCode, String cou
...
+			}
+			return ((!input.category_code.equals(incomingChanges.category_code)
+					&& !incomingChanges.category_code.equals("MISC"))
+					|| (!input.country_code.equals(incomingChanges.country_code)
+					&& !incomingChanges.country_code.equals("XX")));
+		}));
+
+		if (!existingUrlsToUpdate.isEmpty()) {
+			Url.saveAll(Lists.newArrayList(existingUrlsToUpdate));
+		}
```
hellais, line 110: The code here presents a few issues:
1. At the moment in time in which a certain measurements was run the URL had that category code and country code, so by updating it we are in an unexpected way changing past results.
2. In `resultUrlsMap` you are building a map that is keyed on the URL and using this to lookup if the URL requires updating. This creates an undefined behaviour if the same URL is present in two different country lists, because it would share the same key, yet they are two distinct entries in the test list.

OTOH if we don't update them, since the lookup is done based on the url value, we run the risk of displaying URL entries that are mis-categorised.

I don't really see a way around this, beyond keeping a reference to the URL object throughout the testing lifecycle, so you can use that instead of having to redo a lookup at the end of a test run or better yet just merging the `url` and `measurement` tables together as discussed in: https://github.com/ooni/probe/issues/2024.

Maybe since both of these options have tradeoffs which are more or less equivalent, we should go for the solution that results in the simplest code and hence drop all this logic?

app/src/main/java/org/openobservatory/ooniprobe/model/database/Url.java
```diff
@@ -56,6 +75,52 @@ public static Url checkExistingUrl(String input, String categoryCode, String cou
...
+
+		List<Url> existingUrlsToUpdate = Lists.newArrayList(Iterables.filter(existingUrls, input -> {
+			Url incomingChanges = resultUrlsMap.get(input.url);
+			if (incomingChanges == null) {
+				return false;
+			}
+			return ((!input.category_code.equals(incomingChanges.category_code)
+					&& !incomingChanges.category_code.equals("MISC"))
+					|| (!input.country_code.equals(incomingChanges.country_code)
+					&& !incomingChanges.country_code.equals("XX")));
```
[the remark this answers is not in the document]
aanorbel: Thats just part of the existing logic from the `Url` class.
See https://github.com/ooni/probe-android/blob/master/app/src/main/java/org/openobservatory/ooniprobe/model/database/Url.java#L50-L51

app/src/main/java/org/openobservatory/ooniprobe/model/database/Url.java
```diff
@@ -56,6 +75,52 @@ public static Url checkExistingUrl(String input, String categoryCode, String cou
...
+		}));
+
+		if (!existingUrlsToUpdate.isEmpty()) {
+			Url.saveAll(Lists.newArrayList(existingUrlsToUpdate));
+		}
+
+		if (existingUrls.size() != resultUrlsMap.size()) {
+
+			List<String> existingUrlStrings = Lists.newArrayList(Iterables.transform(existingUrls, input -> input.url));
```
hellais, line 114: It's going to be a bit faster when you do the lookup below if you store the existingUrls inside of a `Map`.

It's probably going to be just as fast and perhaps a bit more legible if instead of filtering through the maps you use a `HashMap` and do a set difference between the two.

I am thinking of something like:
```
Set<String> existingUrlSet = new HashSet<String>(Lists.newArrayList(Iterables.transform(existingUrls, input -> input.url)))
Set<String> newUrlSet = new HashSet<string>(resultUrlsMap.values())
newUrlSet.removeAll(existingUrlSet)
```

app/src/main/java/org/openobservatory/ooniprobe/model/database/Url.java
aanorbel: I would go for merging the tables. This would obviously need some code refactoring where necessary. 
For this to happen, the `probe-cli` would have to be updated in turn.

If we want to wait for that to happen, we can halt this pull request for that to happen. Otherwise we can ship what we have and when the `probe-cli` is made to support the table merge, we update the apps appropriately

hellais: Yeah I think we should do it incrementally as the change in probe-cli is going to require quite a bit more work. I suggest we ship this as-is then.

bassosimone (approved):
🐳

bassosimone:
(Here's a reminder that the CI is still in a sorry state: https://github.com/ooni/probe-android/runs/5307420992?check_suite_focus=true#step:4:341)

PR #487, opened 2022-02-17
Display total "Time remaining counter" instead of single timer for single test

aanorbel:
Fixes ooni/probe#1936

## Proposed Changes

  - Set `TestRunBroadRequestReceiver.runtime` to summation of `TestAsyncTask.testSuites` runtime.
  - Set `progress` max value to equal `previousTestsRuntime + currentTestProgress`
  - Calculate `timeLeft` from new values available.

| .  |. | . |
| ------------- | ------------- |  ------------- | 
| Screenshot_20220217_142916  | Screenshot_20220217_143358  | Screenshot_20220217_143413  |


app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
```diff
@@ -174,19 +171,20 @@ protected void onResume() {
             return;
         }
         IntentFilter filter = new IntentFilter("org.openobservatory.ooniprobe.activity.RunningActivity");
-        receiver = new TestRunBroadRequestReceiver();
+        receiver = new TestRunBroadRequestReceiver(preferenceManager, new TestRunnerEventListener());
         LocalBroadcastManager.getInstance(this).registerReceiver(receiver, filter);
         //Bind the RunTestService
         Intent intent = new Intent(this, RunTestService.class);
-        bindService(intent, this, Context.BIND_AUTO_CREATE);
+//        bindService(intent, this, Context.BIND_AUTO_CREATE);
```
bassosimone, line 178:
```suggestion
```

I suggest to just remove this commented out line

app/src/main/java/org/openobservatory/ooniprobe/fragment/ProgressFragment.java
```diff
@@ -90,29 +86,35 @@ public void onResume() {
...
+        if (((Application)getActivity().getApplication()).isTestRunning()){
+
+            progress.setIndeterminate(true);
         if (service != null && service.task != null){
             if (service.task.currentSuite != null)
-                progress.setMax(service.task.currentSuite.getTestList(preferenceManager).length * 100);
+            progress.setMax((int)Stats.of(Lists.transform(
+                    service.task.testSuites,
+                    input -> input.getTestList(preferenceManager).length * 100
+            )).sum());
```
bassosimone, line 105: Is this line's indent correct?

app/src/main/java/org/openobservatory/ooniprobe/fragment/ProgressFragment.java
```diff
@@ -90,29 +86,35 @@ public void onResume() {
...
             if (service.task.currentTest != null)
                 name.setText(getString(service.task.currentTest.getLabelResId()));
         }
+        }
```
bassosimone, line 109: Is this line's indent correct?

app/src/main/java/org/openobservatory/ooniprobe/fragment/ProgressFragment.java
```diff
@@ -123,53 +125,44 @@ public void onDestroy() {
...
-            }
+        public void onProgress(int state, double eta) {
+            if (progress.isIndeterminate())
+                updateUI(receiver.service);
+            progress.setIndeterminate(false);
+            progress.setProgress(state);
+        }
+
+        @Override
+        public void onLog(String value) {}
```
bassosimone, line 148:
```suggestion
        public void onLog(String value) {
            /* nothing */
        }
```

Let's make it explicit that the callback is empty on purpose!

app/src/main/java/org/openobservatory/ooniprobe/fragment/ProgressFragment.java
```diff
@@ -123,53 +125,44 @@ public void onDestroy() {
...
+
+        @Override
+        public void onError(String value) {}
```
bassosimone, line 151:
```suggestion
        public void onError(String value) {
            /* nothing */
        }
```

Let's make it explicit that the callback is empty on purpose!

app/src/main/java/org/openobservatory/ooniprobe/receiver/TestRunBroadRequestReceiver.java
```diff
@@ -0,0 +1,122 @@
+package org.openobservatory.ooniprobe.receiver;
+
```
bassosimone, line 2: Please add some short documentation for this new package!

app/src/main/java/org/openobservatory/ooniprobe/receiver/TestRunBroadRequestReceiver.java
```diff
@@ -0,0 +1,122 @@
...
+import com.google.common.math.Stats;
+
+import org.openobservatory.ooniprobe.common.PreferenceManager;
+import org.openobservatory.ooniprobe.common.service.RunTestService;
+import org.openobservatory.ooniprobe.test.TestAsyncTask;
+import org.openobservatory.ooniprobe.test.suite.AbstractSuite;
+
+import java.util.List;
+
+public class TestRunBroadRequestReceiver extends BroadcastReceiver implements ServiceConnection {
```
bassosimone, line 21: Please add some short documentation for this class.

app/src/main/java/org/openobservatory/ooniprobe/receiver/TestRunBroadRequestReceiver.java
```diff
@@ -0,0 +1,122 @@
...
+        String value = intent.getStringExtra("value");
+        switch (key) {
+            case TestAsyncTask.START:
+                listener.onStart(service);
+                break;
+            case TestAsyncTask.RUN:
+                listener.onRun(value);
+                break;
+            case TestAsyncTask.PRG:
+                List<AbstractSuite> previousTestSuits =
```
bassosimone, line 45: I think the correct naming for the variable should be `previousTestSuites` (i.e., suites not suits)

app/src/main/java/org/openobservatory/ooniprobe/receiver/TestRunBroadRequestReceiver.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Override
+    public void onServiceConnected(ComponentName cname, IBinder binder) {
+        //Bind the service to this activity
+        RunTestService.TestBinder b = (RunTestService.TestBinder) binder;
+        service = b.getService();
+        isBound = true;
+        listener.onStart(service);
+        runtime = (int) Stats.of(Lists.transform(service.task.testSuites, input -> input.getRuntime(preferenceManager))).sum();
+
```
bassosimone, line 88:
```suggestion
```

app/src/main/java/org/openobservatory/ooniprobe/receiver/TestRunBroadRequestReceiver.java
```diff
@@ -0,0 +1,122 @@
...
+    public boolean isBound() {
+        return isBound;
+    }
+
+    public void setBound(boolean bound) {
+        isBound = bound;
+    }
+
+    public interface EventListener {
```
bassosimone, line 105: Please add documentation for this public interface. While I am not an advocate to documenting all methods by default, I still think it would be useful in this case to document what methods of this interface do.

app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
```diff
@@ -266,13 +208,58 @@ private void testEnded(Context context) {
     @Override
     public void onConfirmation(Serializable serializable, int i) {
         if (i == DialogInterface.BUTTON_POSITIVE) {
-            if (service != null)
-                service.task.interrupt();
+            if (receiver.service != null)
+                receiver.service.task.interrupt();
```
bassosimone, line 212: I am not super happy that we're navigating `receiver.service.task.interrupt` and it would perhaps feel a bit more tidy if we could have a method that does the `.service.task.interrupt()` call for us

app/src/main/java/org/openobservatory/ooniprobe/receiver/TestRunBroadRequestReceiver.java
```diff
@@ -0,0 +1,122 @@
...
+import org.openobservatory.ooniprobe.common.service.RunTestService;
+import org.openobservatory.ooniprobe.test.TestAsyncTask;
+import org.openobservatory.ooniprobe.test.suite.AbstractSuite;
+
+import java.util.List;
+
+public class TestRunBroadRequestReceiver extends BroadcastReceiver implements ServiceConnection {
+    private final EventListener listener;
+    private final PreferenceManager preferenceManager;
+    public RunTestService service;
```
bassosimone, line 24: I think the diff would feel a bit more encapsulated if we'd make this field private. (Of course to do this we would need to factor the `.service.task.interrupt()` call as I suggested above.)

app/src/main/java/org/openobservatory/ooniprobe/receiver/TestRunBroadRequestReceiver.java
```diff
@@ -0,0 +1,122 @@
...
+                break;
+            case TestAsyncTask.END:
+                listener.onEnd(context);
+                break;
+        }
+    }
+
+    @Override
+    public void onServiceConnected(ComponentName cname, IBinder binder) {
+        //Bind the service to this activity
```
bassosimone, line 82:
```suggestion
        //Bind the service to this activity
```

Is this comment still relevant? This class does not seem to be an activity anymore.

app/src/main/java/org/openobservatory/ooniprobe/receiver/TestRunBroadRequestReceiver.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    public void setBound(boolean bound) {
+        isBound = bound;
+    }
+
+    public interface EventListener {
+        void onStart(RunTestService service);
+
+        void onRun(String value);
```
bassosimone, line 108: I think the name of this method is not very clear. What is the different in semantics between a "start" event and a "run" event. (I understand you're mainly reshuffling existing code here, but I feel like the previous code was under-documented and I think we should improve a little bit with more docs and better naming.)

app/src/main/java/org/openobservatory/ooniprobe/receiver/TestRunBroadRequestReceiver.java
```diff
@@ -0,0 +1,122 @@
...
+
+        void onProgress(int state, double eta);
+
+        void onLog(String value);
+
+        void onError(String value);
+
+        void onUrl();
+
+        void onInt();
```
bassosimone, line 118: I am not super fond of the name of this method. What does it mean that the "int" event occurs? Does this mean that there is some integer? Or perhaps has this to do with interrupting? In such a case I'll say "onInterrupt".

app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
```diff
@@ -46,7 +44,7 @@
 import localhost.toolkit.app.fragment.ConfirmDialogFragment;
 import localhost.toolkit.app.fragment.MessageDialogFragment;
 
-public class RunningActivity extends AbstractActivity implements ConfirmDialogFragment.OnConfirmedListener, ServiceConnection {
+public class RunningActivity extends AbstractActivity implements ConfirmDialogFragment.OnConfirmedListener {
```
bassosimone, line 47: I'd invest a little more effort and add a one-line of documentation explaining the purpose of this class while we're touching it, such that our future selves more clearly know what this class is about.

app/src/main/java/org/openobservatory/ooniprobe/receiver/TestRunBroadRequestReceiver.java
bassosimone: Also, I am bit confused by the name of this class. Shouldn't it be named `TestRunBroadcastReceiver`? Why is it named `Broad` and not `Broadcast`? Which `Request` is it receiving?

bassosimone (changes requested):
My review could be broadly split into two parts:

1. I have asked for changes that overall seek to improve naming and documentation;

2. I have highlighted a bunch of places where I think we could possibly refactor and make the code a bit more abstract. I think they make sense but my limited Android skills could be failing me and I may be actually providing suggestions that make increase rather than decrease the complexity of the code. In which case, please, let me know and I'll keep in mind the lessons learned today for future reviews.

Thank you for preparing this diff!

app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
```diff
@@ -154,13 +149,15 @@ private void applyUIChanges(){
...
-        runtime = service.task.currentSuite.getRuntime(preferenceManager);
         getWindow().setBackgroundDrawableResource(service.task.currentSuite.getColor());
         if (Build.VERSION.SDK_INT >= 21) {
             getWindow().setStatusBarColor(service.task.currentSuite.getColor());
         }
         animation.setAnimation(service.task.currentSuite.getAnim());
-        progress.setMax(service.task.currentSuite.getTestList(preferenceManager).length * 100);
+        progress.setMax((int)Stats.of(Lists.transform(
+                service.task.testSuites,
+                input -> input.getTestList(preferenceManager).length * 100
```
hellais, line 159: I would suggest renaming the argument for the lambda to `ts` to indicate that it's a testSuite. 

Initially I was confused because I thought we were counting inputs here, but we are actually counting test names. That is to say that when you call `getTestList().length` on the `Websites` testSuite you will get 1, while when you call the same on InstantMessagging you will get 4.

Question: does this have an impact on the progress bar moving slower than it should on the websites test group?

hellais:
The PR looks good to me. I did some light testing of it and I noticed the following two issues (that I believe are not related to this PR though):

When in dark mode the experimental icon has the wrong background color:

Screenshot 2022-02-21 at 12 08 27

When in dark mode the status bar has the wrong color:

Screenshot 2022-02-21 at 12 08 22

aanorbel:
Hi @hellais , I will check the coloring for the progress bar.

app/src/main/java/org/openobservatory/ooniprobe/receiver/TestRunBroadRequestReceiver.java
aanorbel: I was not aware of the full name so i had to make quick work of the situation

app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
```diff
@@ -154,13 +149,15 @@ private void applyUIChanges(){
...
+        )).sum());
```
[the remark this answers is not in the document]
aanorbel: The operation to add component parts is repeated. I would look for a way do remove possible duplications

app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
```diff
@@ -174,19 +171,20 @@ protected void onResume() {
...
 
     @Override
     protected void onPause() {
         super.onPause();
-        if (isBound) {
-            unbindService(this);
-            isBound = false;
+        if (receiver.isBound()) {
+            unbindService(receiver);
+            receiver.setBound(false);
```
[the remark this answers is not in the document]
aanorbel: this would mean passing a context object to the `receiver`  so it can be able to call `unbindService(connection)` once the checks are performed.
That did not strike me as the way to handle the situation. passing an additional argument to the class just to be used in one case for convenience.

app/src/main/java/org/openobservatory/ooniprobe/receiver/TestRunBroadRequestReceiver.java
aanorbel: package level docs not possible without `package-info.java`

aanorbel:
> Screenshot 2022-02-21 at 12 08 27
> 
Hello @hellais , this problem comes specifically from [ExperimentalSuite()](https://github.com/ooni/probe-android/blob/3373ba47fdb9963d0aa7e3b9463c80fffe6220cf/app/src/main/java/org/openobservatory/ooniprobe/test/suite/ExperimentalSuite.java#L25) and [anim/experimental.json](https://github.com/ooni/probe-android/blob/3373ba47fdb9963d0aa7e3b9463c80fffe6220cf/app/src/main/assets/anim/experimental.json#L1) which is not theme aware. I would have to make various changes at a later time to make the resource change with theme.

It is worth noting that this problem exists in all previous versions of the app.

hellais:
Ok I documented the animation background color issue here: https://github.com/ooni/probe/issues/2038. For the rest we can proceed with merging this!

aanorbel:
Sure

app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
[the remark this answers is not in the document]
bassosimone: ✔️

bassosimone (approved):
🐳

PR #496, opened 2022-03-07
fix crash for website re-run

aanorbel:
Fixes https://github.com/ooni/probe/issues/2043

## Proposed Changes

  - Wrap access to `RunTestService` with null check

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/receiver/TestRunBroadRequestReceiver.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/receiver/TestRunBroadRequestReceiver.java
@@ -56,24 +56,26 @@ public class TestRunBroadRequestReceiver extends BroadcastReceiver implements Se
                 listener.onRun(value);
                 break;
             case TestAsyncTask.PRG:
-                List<AbstractSuite> previousTestSuites =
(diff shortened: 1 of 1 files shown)


bassosimone (changes requested):
This diff looks good and intuitive to me: because the service may have become `null` in the meanwhile, we're not using it. I think, however, I may require some clarification on the sequence of events before signing off these changes.

Specifically, I am a bit confused at this time about what the `PRG` event does. I think this is because `PRG` is a very cryptic acronym, so I cannot recall on top of my head what it is.

My second doubt, instead, is how are we going to re-run the test if we don't have a service that we can call?

Thank you!

aanorbel:
Hi @bassosimone , the service is only available after `TestRunBroadRequestReceiver.onServiceConnected` is called. This means the `BroadcastReceiver` can publish a progress when the `ServiceConnection` has not been establised.

bassosimone:
@aanorbel thanks! It has now finally occurred to me that `PRG` is progress! (Can we perhaps rename these defines in a better way so that it's more obvious what they refer to?)

bassosimone (approved):
Everything is clear now after your explanation, thanks a lot!

LGTM 🐳

PR #497, opened 2022-03-07
Do not display VPN warnings when data sharing is disabled

aanorbel:
Fixes https://github.com/ooni/probe/issues/2015

## Proposed Changes

  - Update service `RunningActivity` to only show modal when `VPN` and `upload_results` are enabled.
  - Show `VPN` warning modal if any measurement to be uploaded has `VPN` enabled in annotations. 
 
| Light Theme      | Dark Theme |
| ----------- | ----------- |
|    Screenshot_20220307_215236   |     Screenshot_20220307_215207   |
|   Screenshot_20220307_215251 |   Screenshot_20220307_215326 |
|  Screenshot_20220307_215433 | Screenshot_20220307_215459 |

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
@@ -242,7 +242,10 @@ public class MeasurementDetailActivity extends AbstractActivity implements Confi
     }
 
     private void runAsyncTask() {
-        new ResubmitAsyncTask(this, pm.getProxyURL()).execute(null, measurement.id);
+        measurementsManager.syncMeasurements(this, () -> {
+            new ResubmitAsyncTask(this, pm.getProxyURL()).execute(null, measurement.id);
+            this.load();
+        }, this::load, null, measurement.id);
     }
 
     private void load() {
@@ -265,7 +268,7 @@ public class MeasurementDetailActivity extends AbstractActivity implements Confi
                 Intent share = new Intent(Intent.ACTION_SEND);
                 share.putExtra(Intent.EXTRA_TEXT, measurementsManager.getExplorerUrl(measurement));
                 share.setType("text/plain");
-                Intent shareIntent = Intent. createChooser(share, null);
+                Intent shareIntent = Intent.createChooser(share, null);
                 startActivity(shareIntent);
                 return true;
             default:
--- a/app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
@@ -88,7 +88,8 @@ public class RunningActivity extends AbstractActivity implements ConfirmDialogFr
      */
     public static void runAsForegroundService(AbstractActivity context,
                                               ArrayList<AbstractSuite> testSuites,
-                                              OnTestServiceStartedListener onTestServiceStartedListener) {
+                                              OnTestServiceStartedListener onTestServiceStartedListener,
+                                              PreferenceManager _preferenceManager) {
         if (ReachabilityManager.getNetworkType(context).equals(ReachabilityManager.NO_INTERNET)) {
             new MessageDialogFragment.Builder()
                     .withTitle(context.getString(R.string.Modal_Error))
@@ -99,7 +100,7 @@ public class RunningActivity extends AbstractActivity implements ConfirmDialogFr
                     .withTitle(context.getString(R.string.Modal_Error))
                     .withMessage(context.getString(R.string.Modal_Error_TestAlreadyRunning))
                     .build().show(context.getSupportFragmentManager(), null);
-        } else if (ReachabilityManager.isVPNinUse(context)) {
+        } else if (ReachabilityManager.isVPNinUse(context) && _preferenceManager.isUploadResults()) {
             new AlertDialog.Builder(context, R.style.MaterialAlertDialogCustom)
                     .setTitle(context.getString(R.string.Modal_DisableVPN_Title))
                     .setMessage(context.getString(R.string.Modal_DisableVPN_Message))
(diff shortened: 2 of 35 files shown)


app/src/main/java/org/openobservatory/ooniprobe/model/database/Measurement.java
```diff
@@ -330,4 +330,13 @@ public void setReRun(Context c){
 		this.is_rerun = true;
 		this.save();
 	}
+
+	public class DataRoot {
```
bassosimone, line 334: Please document the purpose of this class! (Just a one-liner is fine)

app/src/main/java/org/openobservatory/ooniprobe/activity/MeasurementDetailActivity.java
```diff
@@ -265,7 +268,7 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 Intent share = new Intent(Intent.ACTION_SEND);
                 share.putExtra(Intent.EXTRA_TEXT, measurementsManager.getExplorerUrl(measurement));
                 share.setType("text/plain");
-                Intent shareIntent = Intent. createChooser(share, null);
+                Intent shareIntent = Intent.createChooser(share, null);
```
bassosimone, line 271: 🙏

app/src/main/java/org/openobservatory/ooniprobe/activity/RunningActivity.java
```diff
@@ -88,7 +88,8 @@
      */
     public static void runAsForegroundService(AbstractActivity context,
                                               ArrayList<AbstractSuite> testSuites,
-                                              OnTestServiceStartedListener onTestServiceStartedListener) {
+                                              OnTestServiceStartedListener onTestServiceStartedListener,
+                                              PreferenceManager _preferenceManager) {
```
bassosimone, line 92: Why does the name here starts with underscore?

app/src/main/java/org/openobservatory/ooniprobe/domain/MeasurementsManager.java
```diff
@@ -109,6 +117,57 @@ public String getReadableEntry(Measurement measurement) throws IOException {
...
+                    .setTitle(activity.getString(R.string.Modal_UploadVPNResults_Title))
+                    .setMessage(activity.getString(R.string.Modal_UploadVPNResults_Message))
+                    .setNegativeButton(R.string.Modal_Cancel, (dialogInterface, i) -> {
+                        dialogInterface.dismiss();
+                        onCancelResubmission.onCancel();
+                    })
+                    .setPositiveButton(R.string.Modal_OK, (dialogInterface, i) -> {
+                        onStartResubmission.onResubmit();
+                    })
+                    .show();
```
bassosimone, line 136: Is there a way to factor this piece of code which seems repeated ~equally more than once?

app/src/main/java/org/openobservatory/ooniprobe/domain/MeasurementsManager.java
```diff
@@ -109,6 +117,57 @@ public String getReadableEntry(Measurement measurement) throws IOException {
...
+        List<Measurement> measurements = Measurement.withReport(context, msmQuery);
+
+        return Lists.transform(measurements, measurement -> {
+            try {
+                return ((Application) context.getApplicationContext()).getGson().fromJson(
+                        FileUtils.readFileToString(
+                                Measurement.getEntryFile(context, measurement.id, measurement.test_name),
+                                StandardCharsets.UTF_8
+                        ),
+                        Measurement.DataRoot.class
```
bassosimone, line 162: I am not getting why `DataRoot` is named as such? Is it named as such because it's the toplevel object representing a measurement?

app/src/main/java/org/openobservatory/ooniprobe/fragment/DashboardFragment.java
```diff
@@ -108,7 +118,8 @@ private void onTestServiceStartedListener() {
                 RunningActivity.runAsForegroundService(
                         (AbstractActivity) getActivity(),
                         testSuite.asArray(),
-                        this::onTestServiceStartedListener
+                        this::onTestServiceStartedListener,
+						preferenceManager
```
bassosimone, line 122: Shouldn't this line be indented like the above line?

bassosimone (commented):
I am a bit confused by the overall objective of this diff and I think I will need some clarification before continuing my review. As far as I can tell, my confusion does not stem directly from the PR itself, but from the original issue.

Here we're not showing a warning when running a measurement over a VPN if uploading results is disabled. Then we show the warning when the user tries to upload measurements previously collected with a VPN enabled.

I think we do not actually want measurements collected over a VPN in the first place. We don't care about them.

Now, suppose I am a user and I don't know that OONI does not want me to run measurements over a VPN. Say I also have disabled data sharing. Then, I would run measurements without issues. But, when I try to submit them, the app tells me I cannot because I was using a VPN. And I would not be amused. I mean, why didn't the app tell me in advance that I was doing work that is not useful to OONI and cannot be submitted? Why did I waste some of my bandwidth and some of my time running those measurements?

Shouldn't the app behave in exactly the opposite way? With this I mean: prevent measurements to be run over a VPN but allow submitting them using a VPN?

I am starting to wonder whether I'm missing something fundamental here?

agrabeli (commented):
Thank you @aanorbel for working on this! I only have feedback on the copy.

Perhaps edit the copy to the following:

Title: Some measurements were collected over a VPN

Paragraph:

If you upload measurements collected while a VPN was enabled, the test results may appear to come from the wrong country.

app/src/main/java/org/openobservatory/ooniprobe/domain/MeasurementsManager.java
lorenzoPrimi: @bassosimone this is a normal alert, where else is it repeated?

PR #511, opened 2022-04-18
 Paginate scrolling in test result screen

aanorbel:
Fixes https://github.com/ooni/probe/issues/1430

## Proposed Changes

  - Introduce Pagging to use [Android Pager](https://developer.android.com/topic/libraries/architecture/paging/v3-overview)
  - Move away from [`HeterogeneousRecyclerAdapter`](https://github.com/xanscale/LocalhostToolkit/blob/master/app/src/main/java/localhost/toolkit/widget/recyclerview/HeterogeneousRecyclerAdapter.java)
  - Depricated classes no longer required

### Sample Pager logs
```
V/FlowLog: Executing query: SELECT * FROM `Result` ORDER BY `start_time` DESC LIMIT 20 OFFSET 0 
V/FlowLog: Executing query: SELECT * FROM `Result` ORDER BY `start_time` DESC LIMIT 20 OFFSET 20 
```


app/src/main/java/org/openobservatory/ooniprobe/adapters/ResultListAdapter.java
```diff
@@ -0,0 +1,298 @@
...
+                return 1;
+            case MiddleBoxesSuite.NAME:
+                return 2;
+            case PerformanceSuite.NAME:
+                return 3;
+            case CircumventionSuite.NAME:
+                return 4;
+            case ExperimentalSuite.NAME:
+            default:
+                return 5;
```
hellais, line 251: These could be defined as ENUMs or const, so that you aren't doing switch casing on opaque numbers.

app/src/main/java/org/openobservatory/ooniprobe/domain/QueryDataSource.java
```diff
@@ -0,0 +1,63 @@
...
+        SQLOperator[] conditions = (testGroupNameFilter != null && !testGroupNameFilter.isEmpty())
+                ? new SQLOperator[]{Result_Table.test_group_name.is(testGroupNameFilter)}
+                : new SQLOperator[0];
+
+        List<Result> response = SQLite.select().from(Result.class)
+                .where(conditions)
+                .limit(20)
+                .offset(pageNumber * 20)
+                .orderBy(Result_Table.start_time, false)
+                .queryList();
```
hellais, line 47: If while you have fetched the first page of data, another result is written to the results table (because there is a test currently running), fetching the next page will be off by one.

You could instead use the `result_id` as a pagination handler and query `WHERE result_id < $last_result_id` to get the next page worth of data.

app/src/main/java/org/openobservatory/ooniprobe/domain/QueryDataSource.java
```diff
@@ -33,26 +38,31 @@ public Integer getRefreshKey(@NonNull PagingState<Integer, Result> pagingState)
...
+                .orderBy(Result_Table.id, false)
                 .queryList();
 
-        // Since 0 is the lowest page number, return null to signify no more pages should
-        // be loaded before it.
-        Integer prevKey = (pageNumber > 0) ? pageNumber - 1 : null;
-
+        Integer prevKey;
+        try {
+            prevKey = (!response.isEmpty()) ? Objects.requireNonNull(Iterables.getFirst(response, null)).id - 20 : null;
```
hellais, line 58: Do you actually need the previous key? In which circumstances are you using this?

I don't think you can derive this by just subtracting 20, because some test results may have been removed and so the list of IDs may have "holes" in it.

app/src/main/java/org/openobservatory/ooniprobe/domain/QueryDataSource.java
```diff
@@ -0,0 +1,73 @@
...
+        } catch (Exception e) {
+            prevKey = null;
+            ThirdPartyServices.logException(e);
+        }
+        // This API defines that it's out of data when a page returns empty. When out of
+        // data, we return `null` to signify no more pages should be loaded
+        Integer nextKey = (!response.isEmpty()) ? Iterables.getLast(response).id : null;
```
bassosimone, line 65:
```suggestion
        Integer nextKey = (!response.isEmpty()) ? Iterables.getLast(response).id + 1 : null;
```

Conceptually, the next page seems to start at index `$last + 1` rather than at index `$last`. That said, I can see how smooth scrolling means that probably we want to keep the last element in scope.

app/src/main/java/org/openobservatory/ooniprobe/domain/QueryDataSource.java
```diff
@@ -0,0 +1,73 @@
...
+    @Override
+    public LoadResult.Page<Integer, Result> load(@NonNull LoadParams<Integer> loadParams, @NonNull Continuation<? super LoadResult<Integer, Result>> continuation) {
+        // Key may be null during a refresh, if no explicit key is passed into Pager
+        // construction. Use 0 as default, because our API is indexed started at index 0
+        int highestId = loadParams.getKey() != null ? loadParams.getKey() : 0;
+
+        List<SQLOperator> conditions = (testGroupNameFilter != null && !testGroupNameFilter.isEmpty())
+                ? Collections.singletonList(Result_Table.test_group_name.is(testGroupNameFilter))
+                : new ArrayList<>();
+        if (highestId != 0) {
```
bassosimone, line 46:
```suggestion
        if (highestId > 0) {
```

This makes the code more robust in case the `highestId` ends up being negative

app/src/main/java/org/openobservatory/ooniprobe/domain/QueryDataSource.java
bassosimone: We need the previous key to return it and allow navigation, IIUC the code.

I don't know how the code behaves when there are holes. It may be that, while you are scrolling back, the Android code will continue to query until it has enough elements to fill the screen, so it may be that we don't need to be more precise at the cost of executing more queries.

I think @aanorbel should clarify this.

aanorbel: To optimize the load time and memory usage, the records loaded will not always stay in memory, so the previous key is used when the records previously loaded are no longet in memory.

app/src/main/java/org/openobservatory/ooniprobe/domain/QueryDataSource.java
aanorbel: The query mechanism has been updated to use `result_id` based pagination

app/src/main/java/org/openobservatory/ooniprobe/domain/QueryDataSource.java
aanorbel: The operand used for the query is `lessThan` so using `Iterables.getLast(response).id + 1`  will mean we lose a record.

aanorbel:
Stale

PR #512, opened 2022-04-19
 when you attempt to re-run a failed test while a test is running the row is deleted

aanorbel:
Fixes https://github.com/ooni/probe/issues/2048

Related to https://github.com/ooni/probe/issues/1709

## Proposed Changes

  - Delete Log file immediately test is re-run and not before.

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/fragment/measurement/FailedFragment.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/fragment/measurement/FailedFragment.java
@@ -9,11 +9,11 @@ import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.fragment.app.Fragment;
 
-import org.openobservatory.ooniprobe.R;
 import org.openobservatory.ooniprobe.activity.AbstractActivity;
 import org.openobservatory.ooniprobe.activity.RunningActivity;
 import org.openobservatory.ooniprobe.common.PreferenceManager;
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/openobservatory/ooniprobe/fragment/measurement/FailedFragment.java
```diff
@@ -56,12 +54,12 @@ public static FailedFragment newInstance(Measurement measurement) {
...
 		testSuite.setTestList(abstractTest);
 		testSuite.setResult(failedMeasurement.result);
-		failedMeasurement.setReRun(getContext());
 
 		RunningActivity.runAsForegroundService((AbstractActivity) getActivity(),
 				testSuite.asArray(),
 				() -> {
 					try {
+						failedMeasurement.setReRun(getContext());
 						getActivity().finish();
```
hellais, line 63: What does `getActivity().finish()` do? Do we perhaps want to mark it as re-run after also that operation is successful?

aanorbel: `getActivity().finish(); ` closes the details page for the item and returns to the list.

`failedMeasurement.setReRun(getContext()); `  marks the measurement as `re-run` among other things.

hellais (approved):
🐳

bassosimone (approved):
🐳

PR #516, opened 2022-06-01
Make it obvious to see all the logs

aanorbel:
Fixes  https://github.com/ooni/probe/issues/1777

## Proposed Changes

  - Include log file appender
  - Fetch and display logs to UI.
  - Add log file sharing.

| . | . | . | . |
|-|-|-|-|
| Screenshot_20220824_185707 | Screenshot_20220824_185741 | Screenshot_20220824_185812 | Screenshot_20220824_185851 |


bassosimone:
This is fantastic! Thank you! I will review ASAP!

bassosimone:
Feature request: possibility of copying the logs! (We're starting to test this!)

bassosimone:
We discussed today with @aanorbel that we'll implement sharing of the whole logfile via Telegram, etc.

bassosimone:
I have used and testing this functionality a bit. Here's some extra feedback:

- [ ] I see some utility in sharing logs as HTML, but I would rather share them as a text file without tags because I think ultimately the presence of tags would only make log processing more complex on our end

- [ ] the "advanced" menu entry now seems to contain two "debug logs" entries, however I think the one that leads to the logs should not be named "debug logs" but perhaps something like "see recent logs"

I tested the sharing feature and it's working as intended, thank you!

app/src/main/AndroidManifest.xml
```diff
@@ -166,6 +178,18 @@
 			android:name=".common.service.RunTestJobService"
 			android:label="Run Test Service"
 			android:permission="android.permission.BIND_JOB_SERVICE"/>
+
+		<provider
+			android:name="androidx.core.content.FileProvider"
+			android:authorities="org.openobservatory.ooniprobe.provider"
+			android:exported="false"
+			android:grantUriPermissions="true">
+			<!-- ressource file to create -->
```
bassosimone, line 187:
```suggestion
			<!-- resource file to create -->
```

app/src/main/java/org/openobservatory/ooniprobe/activity/LogActivity.java
```diff
@@ -0,0 +1,103 @@
...
+import android.view.MenuItem;
+import android.view.View;
+import android.widget.AdapterView;
+import android.widget.ArrayAdapter;
+
+import androidx.core.content.FileProvider;
+import androidx.recyclerview.widget.LinearLayoutManager;
+
+import com.google.common.collect.Lists;
+import com.velmurugan.inapplogger.LogType;
```
bassosimone, line 18: I am confused by the fact that the `app/build.gradle` imports `com.github.aanorbel`'s `inapplogger` while here we seem to be importing the original implementation. Is this intended?

app/src/main/java/org/openobservatory/ooniprobe/activity/LogActivity.java
```diff
@@ -0,0 +1,103 @@
...
+            intentShareFile.setType(URLConnection.guessContentTypeFromName(file.getName()));
+            intentShareFile.putExtra(Intent.EXTRA_STREAM, FileProvider.getUriForFile(
+                            this,
+                            "org.openobservatory.ooniprobe.provider", //(use your app signature + ".provider" )
+                            file)
+                    /*Uri.parse("file://" + file.getAbsolutePath())*/);
+
+            //if you need
+            //intentShareFile.putExtra(Intent.EXTRA_SUBJECT,"Sharing File Subject);
+            //intentShareFile.putExtra(Intent.EXTRA_TEXT, "Sharing File Description");
```
bassosimone, line 91:
```suggestion
```

Let's remember to zap the commented out code before merging

app/src/main/java/org/openobservatory/ooniprobe/common/AppLogger.java
```diff
@@ -0,0 +1,57 @@
+package org.openobservatory.ooniprobe.common;
+
+import android.content.Context;
+
+import androidx.annotation.NonNull;
+
+import com.velmurugan.inapplogger.InAppLogger;
```
bassosimone, line 7: Same question as above

app/src/main/java/org/openobservatory/ooniprobe/common/AppLogger.java
```diff
@@ -0,0 +1,57 @@
...
+    public AppLogger(@NonNull Context context) {
+        logger = new InAppLogger(context);
+        /*TODO(aanorbel): 01/06/2022  Check log retention policy and clear logs appropriately.
+            logs only exist for an app session. Log file grows with each test run making
+            time to load logs to increase.*/
+        logger.deleteOldLog();
+    }
+
+    public void e(String tag, String message) {
+        logger.e(String.format("%s : %s", tag, message));
```
bassosimone, line 31:
```suggestion
        logger.e(String.format("%s: %s", tag, message));
```

app/src/main/java/org/openobservatory/ooniprobe/common/AppLogger.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public void w(String tag, String message) {
+        logger.w(String.format("%s : %s", tag, message));
```
bassosimone, line 35:
```suggestion
        logger.w(String.format("%s: %s", tag, message));
```

app/src/main/java/org/openobservatory/ooniprobe/common/AppLogger.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public void i(String tag, String message) {
+        logger.i(String.format("%s : %s", tag, message));
```
bassosimone, line 39:
```suggestion
        logger.i(String.format("%s: %s", tag, message));
```

bassosimone (commented):
Thanks you! I have used the code and already provided some feedback out of that. This interim review adds a couple more questions and suggestions!

app/src/main/java/org/openobservatory/ooniprobe/activity/LogActivity.java
aanorbel: i added functionality to the package without changing the package name. 
i had to publish the update under a different package name as the owner has not approved new changes to the package.

bassosimone: ✔️

bassosimone:
I am using the diff and I am happy with it. However, I think it would be better if we only selected the "LOG" events and just emitted their severity and content. Currently we're emitting all MK events, which I think only adds confusion.

aanorbel:
We can always change the type of events we log. Establishing the necessary framework for logging and sharing logs was top priority. I would be removing logs for `MK_EVENTS`.

bassosimone:
> We can always change the type of events we log. Establishing the necessary framework for logging and sharing logs was top priority. I would be removing logs for `MK_EVENTS`.

What I suggest is to only log the `MK_EVENTS` that contain log messages. You can identify them by their key which is `log`. There should be a place in the codebase where we already trap this event for logging. That should suffice.

Thank you! 🙌

bassosimone (approved):
🐳

PR #518, opened 2022-06-03
Only run `torsf` and `vanilla_tor` in unattended/background mode

aanorbel:
Fixes https://github.com/ooni/probe/issues/2102

## Proposed Changes

  - Modified experimental suite to return test list based on run mode.

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/test/suite/ExperimentalSuite.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/test/suite/ExperimentalSuite.java
@@ -41,10 +41,9 @@ public class ExperimentalSuite extends AbstractSuite {
             if (getAutoRun()) {
                 list.add(new Experimental("torsf"));
                 list.add(new Experimental("vanilla_tor"));
-            } else {
-                list.add(new Experimental("stunreachability"));
-                list.add(new Experimental("dnscheck"));
             }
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/openobservatory/ooniprobe/test/suite/ExperimentalSuite.java
```diff
@@ -26,14 +27,34 @@ public ExperimentalSuite() {
...
-            list.add(new Experimental("dnscheck"));
-            list.add(new Experimental("torsf"));
-            list.add(new Experimental("vanilla_tor"));
+            if (getAutoRun()) {
+                list.add(new Experimental("torsf"));
+                list.add(new Experimental("vanilla_tor"));
+            } else {
+                list.add(new Experimental("stunreachability"));
+                list.add(new Experimental("dnscheck"));
+            }
```
bassosimone, line 47:
```suggestion
            if (getAutoRun()) {
                list.add(new Experimental("torsf"));
                list.add(new Experimental("vanilla_tor"));
            }
            list.add(new Experimental("stunreachability"));
            list.add(new Experimental("dnscheck"));
```

We should run all experiments in autorun mode and the just the slow ones otherwise.

bassosimone (changes requested):

PR #519, opened 2022-06-14
Unable to view data for uploaded measurements.

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2140

## Proposed Changes

  - Change API call to fetch data from the backend.
  

Screenshot_20220614_090630

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/domain/MeasurementsManager.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/domain/MeasurementsManager.java
@@ -32,6 +32,7 @@ import okhttp3.Callback;
 import okhttp3.OkHttpClient;
 import okhttp3.Request;
 import okhttp3.Response;
+import okhttp3.ResponseBody;
 
 public class MeasurementsManager {
 
@@ -113,8 +114,8 @@ public class MeasurementsManager {
         //measurement.getUrlString will return null when the measurement is not a web_connectivity
         apiClient.getMeasurement(measurement.report_id, measurement.getUrlString()).enqueue(new GetMeasurementsCallback() {
             @Override
-            public void onSuccess(ApiMeasurement.Result result) {
-                downloadMeasurement(result, callback);
+            public void onSuccess(String result) {
+                callback.onSuccess(jsonPrinter.prettyText(result));
             }
 
             @Override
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/openobservatory/ooniprobe/domain/MeasurementsManager.java
```diff
@@ -113,8 +114,8 @@ public void downloadReport(Measurement measurement, DomainCallback<String> callb
         //measurement.getUrlString will return null when the measurement is not a web_connectivity
         apiClient.getMeasurement(measurement.report_id, measurement.getUrlString()).enqueue(new GetMeasurementsCallback() {
             @Override
-            public void onSuccess(ApiMeasurement.Result result) {
-                downloadMeasurement(result, callback);
```
hellais, line 117: Can we get rid of the `downloadMeasurement` method now that we download the measurement directly and don't have to proxy it through the `/measurements` query?

app/src/main/java/org/openobservatory/ooniprobe/domain/callback/GetMeasurementsCallback.java
```diff
@@ -3,28 +3,35 @@
...
-public abstract class GetMeasurementsCallback implements Callback<ApiMeasurement> {
+public abstract class GetMeasurementsCallback implements Callback<ResponseBody> {
     @Override
-    public void onResponse(Call<ApiMeasurement> call, Response<ApiMeasurement> response) {
+    public void onResponse(Call<ResponseBody> call, Response<ResponseBody> response) {
         if (response.isSuccessful() &&
                 response.body() != null &&
-                response.body().hasResultWithMeasurements()) {
-            onSuccess(response.body().firstResult());
+                response.isSuccessful()) {
```
bassosimone, line 18: This seems wrong: you're testing twice whether `response.isSuccessful()`. We should drop the second check.

bassosimone (changes requested):

app/src/main/java/org/openobservatory/ooniprobe/domain/callback/GetMeasurementsCallback.java
aanorbel: sounds good. let me do that.

bassosimone (approved):
🐳

PR #522, opened 2022-06-20
Performance improvement for app launch

aanorbel:
Related to https://github.com/ooni/probe/issues/2150

## Proposed Changes

  - Differ database related tasks in `Application.onCreate` to the background.


aanorbel:
Before 
```
D/Application: app-on-create: begin
D/Application: app-on-create:      1222 ms, dagger-init
D/Application: app-on-create:      147 ms, flow-init
D/Application: app-on-create:      17 ms, lifecycle-init
D/Application: app-on-create:      2912 ms, measurement-init
D/Application: app-on-create:      6 ms, locale-init
D/Application: app-on-create: end, 4304 ms
```

After 
```
D/Application: app-on-create: begin
D/Application: app-on-create:      1429 ms, dagger-init
D/Application: app-on-create:      112 ms, flow-init
D/Application: app-on-create:      17 ms, lifecycle-init
D/Application: app-on-create:      3 ms, measurement-init
D/Application: app-on-create:      12 ms, locale-init
D/Application: app-on-create: end, 1573 ms
```

app/src/main/java/org/openobservatory/ooniprobe/common/Application.java
```diff
@@ -35,26 +38,35 @@ public class Application extends android.app.Application {
...
-			Measurement.deleteUploadedJsons(this);
-		Measurement.deleteOldLogs(this);
-		ThirdPartyServices.reloadConsents(this);
-
+		timings.addSplit("lifecycle-init");
+		executorService.execute(() -> {
+			if (_preferenceManager.canCallDeleteJson())
+				Measurement.deleteUploadedJsons(Application.this);
+			Measurement.deleteOldLogs(Application.this);
+			ThirdPartyServices.reloadConsents(Application.this);
```
bassosimone, line 63:
```Java
ThirdPartyServices.reloadConsents(Application.this);
```

I am not convinced by this line. For two reasons:

1. are we sure that it's safe to initialize third party services from a background thread while the previous semantics was that the rest of the app _happens after_ we have initialized third party services?
2. what if the current services support this kind of async initialization and everything is WAI _but_ in the future we add a new third-party service behind the opaque `reloadConsents` API and we break it? What ways we have to ensure our future selves would not commit this mistake? Would documentation be enough for this?

Additionally, I think that, conceptually, third-party services are lifecycle-init and not measurement-init.

bassosimone (commented):
You have measured the part that looks critical and have made it run in background to speed up the start-up. The approach looks correct ("measure first").

However, I think we should not consider this work complete without going the extra mile and performing the following actions:

1. <strike>document the performance in cases similar to the original PR (i.e., ~600 results)</strike> (edit: I am sorry but i landed directly on the diff and completely missed https://github.com/ooni/probe-android/pull/522#issuecomment-1160115292);
2. explain why the current algorithm is _so slow_.

My bias towards this algorithm is that it was clearly not designed to scale to hundreds of results (and possibly thousands of measurements) and we should also verify or falsify this opinion of mine while we're trying to address the problem.

It may be that, beyond merely making the app slow at startup, the algorithm we are using is also causing several API calls. It may also _not_ be the case but I think it is worth taking a look to the whole extent of the problem.

But, that we need to background a thread at startup in order to make startup faster seems a bit sweeping under the rug. I argue we should get to the bottom of this issue and have a clear explanation of what is happening. Once we have that, we can then create a follow-up issue for improving the situation and bless this pull request as an hotfix.

(If we're releasing soon, it's also fine to land this PR ~now as long as there's an issue that calls for investigating what is really happening at app startup.)

aanorbel:
Increase in measurements doesn't affect the time taken to schedule background task. 
Optimization would have to be done to improve the time taken for the background task to complete.

```
D/Application: app-on-create: begin
D/Application: app-on-create:      1243 ms, dagger-init
D/Application: app-on-create:      136 ms, flow-init
D/Application: app-on-create:      15 ms, lifecycle-init
D/Application: app-on-create:      1 ms, measurement-init
D/Application: app-on-create:      157 ms, third-party-services-init
D/Application: app-on-create:      10 ms, locale-init
D/Application: app-on-create: end, 1562 ms
```

aanorbel:
Increase in measurements doesn't affect the time taken to schedule background task. 
Optimization would have to be done to improve the time taken for the background task to complete.

```
D/Application: app-on-create: begin
D/Application: app-on-create:      1243 ms, dagger-init
D/Application: app-on-create:      136 ms, flow-init
D/Application: app-on-create:      15 ms, lifecycle-init
D/Application: app-on-create:      1 ms, measurement-init
D/Application: app-on-create:      157 ms, third-party-services-init
D/Application: app-on-create:      10 ms, locale-init
D/Application: app-on-create: end, 1562 ms
```

app/src/main/java/org/openobservatory/ooniprobe/common/Application.java
```diff
@@ -35,26 +38,35 @@ public class Application extends android.app.Application {
...
+		});
+		timings.addSplit("measurement-init");
 		LocaleUtils.setLocale(new Locale(_preferenceManager.getSettingsLanguage()));
 		LocaleUtils.updateConfig(this, getBaseContext().getResources().getConfiguration());
+		timings.addSplit("locale-init");
+		timings.dumpToLog();
```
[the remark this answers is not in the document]
aanorbel: Without Background task
```
D/Application: app-on-create: begin
D/Application: app-on-create:      1234 ms, dagger-init
D/Application: app-on-create:      113 ms, flow-init
D/Application: app-on-create:      16 ms, lifecycle-init
D/Application: app-on-create:      9681 ms, measurement-init
D/Application: app-on-create:      29081 ms, third-party-services-init
D/Application: app-on-create:      16 ms, locale-init
D/Application: app-on-create: end, 40141 ms
```

Background task.
```
D/Application: app-on-create: begin
D/Application: app-on-create:      1243 ms, dagger-init
D/Application: app-on-create:      136 ms, flow-init
D/Application: app-on-create:      15 ms, lifecycle-init
D/Application: app-on-create:      1 ms, measurement-init
D/Application: app-on-create:      157 ms, third-party-services-init
D/Application: app-on-create:      10 ms, locale-init
D/Application: app-on-create: end, 1562 ms
```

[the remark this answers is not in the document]
aanorbel: Timing for `163` measurements.

bassosimone:
> Optimization would have to be done to improve the time taken for the background task to complete.

Yeah! So, I think the best optimization we can implement here is that of dropping the CheckReportID call and instead relying on a different policy regarding the JSONs stored on the device. I already took note of dropping this API from the engine https://github.com/ooni/probe/issues/2161. I think we should also have a similar issue for the mobile apps.

> ```D/Application: app-on-create: begin
> D/Application: app-on-create:      1243 ms, dagger-init```

By the way, I am _also_ very surprised by the time it takes to initialize dagger! 😮

aanorbel:
I saw the numbers for dagger. Its somewhat of a known issue without any trivial solutions. I intend to look deeper into the issue.

aanorbel:
@bassosimone ,@hellias, [Hilt](https://dagger.dev/hilt/) promises an improved runtime performance  in dependencies initialisation. I would like to go ahead and look into this before taking on such a migration.

aanorbel:
After examination, we have decided to drop calls to `check_report_id` and delete json logs immediately measurements are uploaded.

bassosimone:
@aanorbel my understanding is that this PR can be closed because https://github.com/ooni/probe-android/pull/532 exists and addresses the issue in a different way. Is that correct?

aanorbel:
https://github.com/ooni/probe-android/pull/532  will continue to work and optimize the process but will not delete old logs on the system.

app/src/main/java/org/openobservatory/ooniprobe/common/Application.java
```diff
@@ -35,24 +38,39 @@ public class Application extends android.app.Application {
...
 		ProcessLifecycleOwner.get().getLifecycle().addObserver(appLifecycleObserver);
 		Measurement.deleteOldLogs(this);
 		ThirdPartyServices.reloadConsents(this);
 
+		timings.addSplit("lifecycle-init");
+		executorService.execute(() -> {
+			if (_preferenceManager.canCallDeleteJson())
+				Measurement.deleteUploadedJsons(Application.this);
+			Measurement.deleteOldLogs(Application.this);
+		});
```
bassosimone, line 66: I am not completely convinced by running this step in parallel and I may need additional data to understand whether this change is actually thread safe. Is the code that we're calling inside this lambda thread safe or synchronized. If not, we may be causing a bunch of data races in this PR.

bassosimone (commented):
I am okay with the idea of adding support for measuring timing. I am less convinced about the robustness of deleting measurements in parallel without proper synchronization (AFAICT, it may be that I am missing something here).

I would additionally like to know what is the plan with respect to making these metrics that we are collecting actionable. Is the plan that we will run tests on our devices and then use this information to improve the app? Or, is the general idea that we would also log timing information thus allowing users to diagnose and let us know about performance issues _they_ are experiencing?

Thanks!

app/src/main/java/org/openobservatory/ooniprobe/common/Application.java
aanorbel: Hi @bassosimone  this code segment is meant to be run in another thread other than the main thread.
The only thing it does is remove the timing from the main thread. Since no other process is accessing the `logs` and `json` files in the few seconds it uses to delete them, I don't see a problem with syncronization.

bassosimone: Ok, sounds good!

bassosimone:
> I am okay with the idea of adding support for measuring timing. I am less convinced about the robustness of deleting measurements in parallel without proper synchronization (AFAICT, it may be that I am missing something here).

It seems there's no risk of race as explained by @aanorbel in https://github.com/ooni/probe-android/pull/522#discussion_r1115335761

> 
> I would additionally like to know what is the plan with respect to making these metrics that we are collecting actionable. Is the plan that we will run tests on our devices and then use this information to improve the app? Or, is the general idea that we would also log timing information thus allowing users to diagnose and let us know about performance issues _they_ are experiencing?

It seems there's no intention of exposing this information but it's just for us and for debugging

bassosimone (approved):
🐳

PR #523, opened 2022-06-20
Fix: `RunningActivity` back button

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2156
## Proposed Changes

  - Remove code path blocking activity exit and ensure test is still running.


bassosimone (approved):
🐳

PR #532, opened 2022-07-29
update measurement-log retention policy

aanorbel:
Fixes https://github.com/ooni/probe/issues/2162

## Proposed Changes

  - [x] Delete `Measurement` log when measurement cli emits `task_terminated` and measurement has been uploaded.
  - [x] Delete `Measurement` log when measurement is re-submitted.

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/common/ResubmitTask.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/common/ResubmitTask.java
@@ -154,6 +154,9 @@ public class ResubmitTask<A extends AbstractActivity> extends NetworkProgressAsy
                 m.result.load();
                 if(!d.measurementsManager.reSubmit(m, session)){
                     errors++;
+                }else {
+                    m.deleteEntryFile(getActivity());
+                    m.deleteLogFile(getActivity());
                 }
             }
         }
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/openobservatory/ooniprobe/test/test/AbstractTest.java
```diff
@@ -189,7 +189,7 @@ void run(Context c, PreferenceManager pm, Gson gson, Settings settings, Result r
                         setUploaded(true, event.value);
                         break;
                     case "failure.measurement_submission":
-                        setUploaded(false, event.value);
+                        setUploaded(false, event.value );
```
bassosimone, line 192:
```suggestion
                        setUploaded(false, event.value);
```

app/src/main/java/org/openobservatory/ooniprobe/common/ResubmitTask.java
```diff
@@ -152,6 +153,8 @@ protected Boolean doInBackground(Integer... params) {
                 m.result.load();
                 if(!d.measurementsManager.reSubmit(m, session)){
                     errors++;
+                }else {
+                    m.deleteEntryFile(getActivity());
```
bassosimone, line 157: By reading the code inside `Measurement.java`, I see that we have the `hasReportFile` method and then we have `getEntryFile` and `deleteEntryFile`. Could we refactor the last two methods to become `getReportFile` and `deleteReportFile`? I think the "report" based naming is more clear than the "entry" based naming, judging on the terminology I am aware of and I reckon we have been using.

bassosimone (commented):
My understanding based on reading the source code of `Measurement.java` is that this change only impacts the measurements saved on disk. I was initially confused by the pull request title mentioning "log", which is another entity (i.e., the logs generated by nettests while they are running). In light of that, I would recommend updating the PR's title.

bassosimone (approved):
🐳

PR #540, opened 2022-08-22
Fix: Permission Update android 12+

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2243

- [x] Notification not dismissed once tests are completed.
- [x] `JobInfo.Builder.setPersisted` requires `android.permission.RECEIVE_BOOT_COMPLETED`  Update IDE to `Android Studio Electric Eel | 2022.1.1` minimum
- [x] `TestAsyncTask.registerConnChange` requires `android.permission.ACCESS_NETWORK_STATE`  Update IDE to `Android Studio Electric Eel | 2022.1.1` minimum


app/src/main/java/org/openobservatory/ooniprobe/common/service/RunTestService.java
```diff
@@ -211,9 +211,11 @@ public void onReceive(Context context, Intent intent) {
...
                 stopTest();
             }
-            //This is used to close the notification tray
-            Intent it = new Intent(Intent.ACTION_CLOSE_SYSTEM_DIALOGS);
-            context.sendBroadcast(it);
+            // INFO: No need to use ACTION_CLOSE_SYSTEM_DIALOGS in this situation.
+            // That's because, if your app calls startActivity()
+            // while a window is on top of the notification drawer,
+            // the system closes the notification drawer automatically.
+            // https://developer.android.com/about/versions/12/behavior-changes-all?msclkid=3ad37f25cf7411ecb536010741f51e42#close-system-dialogs-exceptions
```
bassosimone, line 218: I would have expected the code to call `sendBroadcast` only when Android is `<= 11` but the code is not doing that. This fact leads me to ask whether the SDK changed such that the system is somehow doing the right thing when we are running on Android `<= 11`. If that's the case, do we need to test that? Did we test that / how can we test that? Also, in such a case, I would recommend extending the text you wrote above (thank you!) to include a mention that the code will be working as intended also for Android `<= 11`.

bassosimone (commented):
I will need some help to understand the new code's behavior with Android `<= 11`.

app/src/main/java/org/openobservatory/ooniprobe/common/service/RunTestService.java
aanorbel: The situation is more like the API has evolved. The default behaviour of `ACTION_CLOSE_SYSTEM_DIALOGS` now needs a permission which is a system permission on all android versions and cannot be granted to an app.

app/src/main/java/org/openobservatory/ooniprobe/common/service/RunTestService.java
```diff
@@ -211,9 +211,11 @@ public void onReceive(Context context, Intent intent) {
             if (action != null && action.equals("interrupt_test")) {
                 stopTest();
             }
-            //This is used to close the notification tray
-            Intent it = new Intent(Intent.ACTION_CLOSE_SYSTEM_DIALOGS);
-            context.sendBroadcast(it);
+            // INFO: No need to use ACTION_CLOSE_SYSTEM_DIALOGS in this situation.
+            // That's because, if your app calls startActivity()
+            // while a window is on top of the notification drawer,
+            // the system closes the notification drawer automatically.
```
bassosimone, line 217:
```suggestion
            // the system closes the notification drawer automatically
            // and this happens on _any_ Android version.
```

app/src/main/java/org/openobservatory/ooniprobe/common/service/RunTestService.java
bassosimone: Gotcha, thanks for explaining!

bassosimone (approved):
🐳

PR #547, opened 2022-11-25
chore: use ooni/probe-cli v3.16.7

bassosimone:


bassosimone:
I am going to merge with broken tests because I have opened an issue about fixing them: https://github.com/ooni/probe/issues/2386

PR #549, opened 2022-12-12
Feat: Add `News Media Scan` flavor

aanorbel:
## Proposed Changes

- Update application based on designs

| Light      | Dark |
| ----------- | ----------- |
| 01_onboarding_2      | .       |
| 01_onboarding_2 | . |
| 02_onboarding_2_q1 | . |
| 02_onboarding_2_q2 | . |
| 03_automated_testing | . |
| 04_crash_report | . |
| 05_default_settings | . |
| 06_dashboard | 06_dashboard | 
| 07_test_results      | 07_test_results |
| 08_test_results_dash | 08_test_results_dash |
| 08_test_results_webconnectivity | 08_test_results_webconnectivity |
| 09_settings |  09_settings |
| 10_info |  10_info |

# Tasks
- [x] Test card coloring.
- [x] Onboarding colours.
- [x] Grainy images.
- [x] About page.


app/build.gradle
```diff
@@ -83,6 +83,14 @@ android {
         full {
             dimension 'license'
         }
+        dw {
+            dimension 'testing'
+            applicationId 'com.dw.ooniprobe'
+            resValue "string", "APP_NAME", "DW Media Probe"
+            buildConfigField 'String', 'BASE_SOFTWARE_NAME', '"dw-media-probe-android"'
+            versionName '0.0.1'
```
bassosimone, line 91: In terms of maintainability, we should figure out before hand a way to map the dw-flavoured-app version to the actual ooniprobe-android version. The simplest scheme that comes to mind is that of using the same version number and the same version code we use for ooniprobe-android. Because using the same version number may be odd (would it be odd?), perhaps we can just share the version code? What do you think?

app/src/dw/AndroidManifest.xml
```diff
@@ -0,0 +1,43 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:tools="http://schemas.android.com/tools">
```
bassosimone, line 2: Did you test whether it's possible to install the dw-flavoured-app and ooniprobe side by side? As far as I know, it is currently not possible to install OONI Probe stable and OONI Probe experimental side by side. While that may be acceptable because it's only us who install the experimental app, not allowing installation side by side will be a bit more confusing for people. That said, also allowing installing side-by-side is weird because we are going to have tow ~ooniprobe apps competing for collecting measurements. Did you spend time thinking about which would be the best way to handle this side by side scenario?

app/src/dw/res/xml/preferences_global.xml
```diff
@@ -0,0 +1,404 @@
+<PreferenceScreen xmlns:android="http://schemas.android.com/apk/res/android"
```
bassosimone, line 1: My understanding is that this file completely duplicates the ooniprobe preferences. I think we should start thinking about a way to be sure we always sync up preferences changes. Did you already start to think about this? What do you think is the best strategy to ensure this happens?

app/src/dw/res/drawable/ooni_logo.xml
```diff
@@ -0,0 +1,30 @@
+<vector xmlns:android="http://schemas.android.com/apk/res/android"
```
bassosimone, line 1: I assume this log is there because we want to say something like "powered by", right?

bassosimone (approved):
I am fine with merging this code into the master branch already under the argument that it's easier to keep things in sync if we have this code in tree.

In this review, I have tried to think about everything that we should clearly define for us to keep this app in sync with out mainline app. So, I think the focus of the discussion should not be much the code you're adding here, which is fine, but rather how we're planning on maintaining it. We can have this discussion as part of merging or we can create an issue describing the aspects we need to discuss in greater detail, depending on what's more convenient to you.

🐳

engine/build.gradle
```diff
@@ -33,6 +36,7 @@ dependencies {
     // For the stable and dev app flavours we're using the library
     // build published at Maven Central.
     stableImplementation 'org.ooni:oonimkall:2022.12.08-092817'
+    dwImplementation 'org.ooni:oonimkall:2022.12.08-092817'
```
bassosimone, line 39: ❗ we need to update this once we merge https://github.com/ooni/probe-android/pull/552

bassosimone (approved):
🐳

hellais:
Issues reported on slack:

1. The app crashed twice. Once when I first opened it and went to settings. The second time when the test finished and I clicked on the results
2. The eta doesn't seem to ever appear. For the full duration of the test it always says calculating ETA
3. The app cards for the disabled tests should not be clickable. Maybe we could have them go to the settings to enable them if we wish to
4. When I go into the settings and enable one of the tests for the groups. It’s then not possible to re-disable them as when I click on them I get an error saying “Please enable at least one test”

aanorbel:
- [ ] The app crashed twice. Once when I first opened it and went to settings. The second time when the test finished and I clicked on the results
- [ ] The eta doesn't seem to ever appear. For the full duration of the test it always says calculating ETA
- [x] The app cards for the disabled tests should not be clickable. Maybe we could have them go to the settings to enable them if we wish to  https://github.com/ooni/probe-android/pull/562
- [x]  When I go into the settings and enable one of the tests for the groups. It’s then not possible to re-disable them as when I click on them I get an error saying “Please enable at least one test” https://github.com/ooni/probe-android/pull/561

PR #553, opened 2023-01-10
Modifies disabled state of test list.

aanorbel:
## Proposed Changes

  - Change the color of disabled cards
  - Charge separator to text `Disabled Tests`
  
|.|.|
|-|-|
| Screenshot_20230110_224838 | Screenshot_20230110_224911 |

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/common/service/ServiceUtil.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/common/service/ServiceUtil.java
@@ -101,7 +101,7 @@ public class ServiceUtil {
         if (suite != null) {
             Intent serviceIntent = new Intent(app, RunTestService.class);
             serviceIntent.putExtra("testSuites", testSuites);
-            serviceIntent.putExtra("storeDB", true);
+            serviceIntent.putExtra("storeDB", false);
             ContextCompat.startForegroundService(app, serviceIntent);
         }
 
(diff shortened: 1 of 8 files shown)


bassosimone (commented):
This PR seems one that @hellais should review, so I'll remove myself as reviewer.

app/src/main/res/values/colors.xml
```diff
@@ -134,4 +134,6 @@
 	<color name="color_cyan7">#1098ad</color>   <!-- not_used -->
 	<color name="color_cyan8">#0c8599</color>   <!-- not_used -->
 	<color name="color_cyan9">#0b7285</color>
+	<color name="disabled_test_text">#cccccc</color>
```
hellais, line 137: This should be a color from our color scheme.

I suggest using #ced4da (gray4)

hellais:
This looks good. I suggest just changing the color of the text in the light mode to a gray that is in our color scheme. Once that is done this is good to go.

app/src/main/java/org/openobservatory/ooniprobe/common/service/ServiceUtil.java
```diff
@@ -101,7 +101,7 @@ public static void callCheckInAPI(Application app) {
         if (suite != null) {
             Intent serviceIntent = new Intent(app, RunTestService.class);
             serviceIntent.putExtra("testSuites", testSuites);
-            serviceIntent.putExtra("storeDB", true);
+            serviceIntent.putExtra("storeDB", false);
```
hellais, line 104: What does this change do?

hellais (approved):
Left a clarifying question, but otherwise LGTM

app/src/main/java/org/openobservatory/ooniprobe/common/service/ServiceUtil.java
aanorbel: This change is linked to the earlier PR https://github.com/ooni/probe-android/pull/543 which changed the value. This corrects that change which was not supposed to be merged.

PR #554, opened 2023-01-11
Add support to OONI Probe for re-testing of an anomalous URL

aanorbel:
Fixes https://github.com/ooni/probe/issues/2367

## Proposed Changes

  - Add `reRun` menu option `MeasurementDetailActivity`  when the Measurement is anomalous.
  - Run test with the single URL that presets the anomaly

|.|.|
|-|-|
| Screenshot_20230112_132910 | Screenshot_20230112_132943|
| Screenshot_20230112_133022 | Screenshot_20230112_132958 |
| Screenshot_20230112_133154 | .|

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/domain/GetTestSuite.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/domain/GetTestSuite.java
@@ -53,4 +53,28 @@ public class GetTestSuite {
 
         return testSuite;
     }
+
+    public AbstractSuite getFrom(Result result, List<String> inputs) {
+        AbstractSuite testSuite = result.getTestSuite();
+        WebConnectivity test = new WebConnectivity();
+
+        // possible NPE from measurements whose url's are null.
+        List<Url> urls = Lists.transform(
+                Lists.newArrayList(
+                        Iterables.filter(result.getMeasurements(), input -> input.url != null)
+                ),
+                measurement -> new Url(
+                        measurement.url.url,
+                        measurement.url.category_code,
+                        measurement.url.country_code
+                )
+        );
+
+        Url.saveOrUpdate(urls);
+
+        test.setInputs(inputs);
+        testSuite.setTestList(test);
+
+        return testSuite;
+    }
 }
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/openobservatory/ooniprobe/domain/GetTestSuite.java
```diff
@@ -53,4 +53,28 @@ public AbstractSuite getFrom(Result result) {
...
     }
+
+    public AbstractSuite getFrom(Result result, List<String> inputs) {
+        AbstractSuite testSuite = result.getTestSuite();
+        WebConnectivity test = new WebConnectivity();
+
+        // possible NPE from measurements whose url's are null.
+        List<Url> urls = Lists.transform(
+                Lists.newArrayList(
+                        Iterables.filter(result.getMeasurements(), input -> input.url != null)
```
bassosimone, line 64: I think here we should _additionally_ filter by the `Measurement.test_name` and ensure we only run this code for Web Connectivity measurements. Because we are also doing something similar in the menu, this code will only be introduced for robustness (which makes sense given that code in different files and packages is likely to diverge sooner or later). As to why I am advocating for this change, the reason is again robustness. Not only we want to avoid trying to re-run experiments with null URLs (which you correctly checked for), but also we'd like to be safe in case we introduce new experiments with non-null URLs in the future.

app/src/main/java/org/openobservatory/ooniprobe/domain/GetTestSuite.java
```diff
@@ -53,4 +53,28 @@ public AbstractSuite getFrom(Result result) {
 
         return testSuite;
     }
+
+    public AbstractSuite getFrom(Result result, List<String> inputs) {
```
bassosimone, line 57: Because this method is only mean to rerun Web Connectivity results, I think we should give this function a more explanatory name. Additionally, we should probably define what should be the expected behavior if feeding this method with a result different from Web Connectivity, and write a test to ensure this is the case.

bassosimone (changes requested):

agrabeli (commented):
Looks good to me. Thank you @aanorbel for working on this! <3

bassosimone (approved):
🐳

PR #555, opened 2023-01-17
RTL support on Experimental Test overview description

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2346

## Proposed Changes

  - Explicitly set text direction for `OverviewActivity` description text view if language is RTL given that the localization is dynamic.

| Before | After |
|-|-|
| Screenshot_20230117_120245 | Screenshot_20230117_120508 |


hellais (approved):
🐳

PR #558, opened 2023-02-23
Fix: Google play priority exception

aanorbel:
## Proposed Changes

  - Add safety check for `NPE` and `RuntimeExceptions


bassosimone (approved):
🐳 💯

PR #561, opened 2023-03-13
Allow all tests under category to be disabled

aanorbel:
Fixes https://github.com/ooni/probe/issues/2437

## Proposed Changes

  - Remove code checks for prevent disable of all categories.

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/fragment/PreferenceFragment.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/fragment/PreferenceFragment.java
@@ -185,9 +185,8 @@ public class PreferenceFragment extends ExtendedPreferenceFragment<PreferenceFra
         }
         else if (preference instanceof SwitchPreferenceCompat) {
             //Call this code only in case of category or tests
-            if (Arrays.asList(getActivity().getResources().getStringArray(R.array.CategoryCodes)).contains(key) ||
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/openobservatory/ooniprobe/fragment/PreferenceFragment.java
```diff
@@ -185,9 +185,8 @@ else if (preference instanceof EditTextPreference) {
         }
         else if (preference instanceof SwitchPreferenceCompat) {
             //Call this code only in case of category or tests
-            if (Arrays.asList(getActivity().getResources().getStringArray(R.array.CategoryCodes)).contains(key) ||
-                    Arrays.asList(getActivity().getResources().getStringArray(R.array.preferenceTestsNames)).contains(key))
-                checkAtLeastOneEnabled(sharedPreferences, key);
+            // if (Arrays.asList(getActivity().getResources().getStringArray(R.array.CategoryCodes)).contains(key) ||
+            //        Arrays.asList(getActivity().getResources().getStringArray(R.array.preferenceTestsNames)).contains(key))
```
bassosimone, line 189: I think we should remove the whole if branch

bassosimone (changes requested):
I think there are three things to do here:

1. create a small reference issue explaining the problem

2. referencing the issue in the PR

3. deleting the whole since branch if we don't need it anymore

bassosimone (approved):
🐳

PR #567, opened 2023-04-18
Fix: RuntimeException: Unable to start service

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2453

## Proposed Changes

  - Simplify the process of running `RunTestService`, fixing failure of automated runs.

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/test/suite/AbstractSuite.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/test/suite/AbstractSuite.java
@@ -133,8 +133,9 @@ public abstract class AbstractSuite implements Serializable {
 		return list;
 	}
 
-	public boolean isTestEmpty() {
-		return testList.length == 0;
+	public boolean isTestEmpty(PreferenceManager preferenceManager) {
+		getTestList(preferenceManager);
+		return testList == null || testList.length == 0;
 	}
 
 	public boolean getAutoRun() {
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/openobservatory/ooniprobe/test/suite/AbstractSuite.java
```diff
@@ -133,8 +133,9 @@ public ArrayList<AbstractSuite> asArray(){
 		return list;
 	}
 
-	public boolean isTestEmpty() {
-		return testList.length == 0;
+	public boolean isTestEmpty(PreferenceManager preferenceManager) {
+		getTestList(preferenceManager);
+		return testList == null || testList.length == 0;
```
bassosimone, line 138: Do I understand it correctly that the bugfix is the above line and the rest of the diff is work to make sure we're always going to call `isTestEmpty` with the preference manager?

Can you explain why do we need to call `getTestList(preferenceManager)` here? Is this a function that mutates the state of the `AbstractSuite`? (I assume `testList` is then an attribute?)

bassosimone (commented):
Thank you! I am asking questions to better understand the diff!

app/src/main/java/org/openobservatory/ooniprobe/test/suite/AbstractSuite.java
aanorbel: `testList` will remain empty until `getTestList(preferenceManager)` is called at least once, allowing the subclass to add the component tests based on preferences to the `testList`.

bassosimone: I see, thank you!

bassosimone (approved):
🐳

PR #538, opened 2022-08-11
Different number of website categories for OONI Probe Mobile and Desktop

aanorbel:
Fixes https://github.com/ooni/probe/issues/2205

## Proposed Changes

  - removes MISC from website categories preference.

PR #569, opened 2023-04-28
Fix: Crash on custom website list test results

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2462

## Proposed Changes

  - Include `MISC` removed in https://github.com/ooni/probe-android/pull/538


bassosimone (approved):
🐳

PR #575, opened 2023-05-23
fix: failing tests and NPE registed when starting `RunTestService`

aanorbel:
## Proposed Changes

  - Update `EventResultFactory.buildLog` to include a log level to FIX resulting `FileNotFoundException`.
  - Update tests filter to remove tests which are null before checking if component tests are empty.


bassosimone (approved):
🐳

PR #580, opened 2023-06-28
Fix: cannot move app to SD-Card

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2019

## Proposed Changes

  - Modified android manifest to allow app installation in sd-card 

Screenshot_20230628_113830


aanorbel:
Evaluate advantages
Read https://android.stackexchange.com/tags/app2sd/info

PR #583, opened 2023-07-11
Chore: Update Gradle and Android core dependencies

aanorbel:


hellais (approved):
In order to get this to work on my local system I had to install a JDK version 17.

I did this by doing:
```
brew install openjdk@17
sudo ln -sfn /usr/local/opt/openjdk@17/libexec/openjdk.jdk /Library/Java/JavaVirtualMachines/openjdk-17.jdk
```
Then I went into the android studio build settings to change the Build Tools -> Gradle settings to point to the new JDK location.

After that it builds and works as intended.

LGTM 🐳

bassosimone (approved):
🐳

PR #584, opened 2023-07-12
feat: support tor and torsf proxies

bassosimone:
Unfortunately, because of https://github.com/ooni/probe/issues/2406, we are going to see crashes when using these proxies.

This diff is part of https://github.com/ooni/probe/issues/2500. Since we're increasingly being blocked, it makes sense to exposes all the possible proxies we can feature.

We're going to touch upon the same files again once we land the https://github.com/ooni/probe-cli/pull/1162 pull request.


aanorbel (approved):
LGTM

aanorbel:
Proxies not functioning. More testing required

bassosimone:
As I realized in https://gitlab.torproject.org/tpo/core/tor/-/issues/40774#note_2959394, if we use tor for circumvention, we cannot use it for measuring, _until_ we use Arti. The issue is that `tor_main` uses static variables, so two concurrent invocations of `tor_main` create spectacular opportunities for data races. In light of this, I am going to close this PR and update the relevant issue. We can always pick this PR up at a later time and adapt it into another PR, when we switch to using Arti.

PR #587, opened 2023-07-21
fix: visual glitch after editing categories

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2506

## Proposed Changes

  - Force activity to recreate after categories are updated.


bassosimone (approved):
🐳

PR #588, opened 2023-07-22
fix: log view crashes when using debug logs

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2508

## Proposed Changes

  - Migrate from `Butterknife` to `ViewBinding` in `PreferenceGlobalFragment`, moving fragment replacement to `onStart` to prevent race condition
  -  Adds safe defaults for instances where logs are empty cased by debug logs.


bassosimone (approved):
🐳

PR #590, opened 2023-07-24
fix(applogger): expose MK_EVENT LOG_DEBUG

bassosimone:
Closes https://github.com/ooni/probe/issues/2508#issuecomment-1647648903


aanorbel (approved):
🐳

PR #623, opened 2023-10-13
Feat: Add support for `http` and `https` proxies

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2561

## Proposed Changes

  - Update support for socks proxy to use dropdown for selecting a scheme.
  - Update code to save and retrieve the various proxies.

Screenshot_20231013_171055

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/common/ProxySettings.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/common/ProxySettings.java
@@ -83,9 +83,9 @@ public class ProxySettings {
         if (protocol == ProxyProtocol.SOCKS5||protocol == ProxyProtocol.HTTP||protocol == ProxyProtocol.HTTPS) {
             // Alright, we now need to construct a new SOCKS5 URL. We are going to defer
             // doing that to the Java standard library (er, the Android stdlib).
-            String urlStr = protocol+"://" + hostname + ":" + port + "/";
+            String urlStr = protocol.getProtocol()+"://" + hostname + ":" + port + "/";
             if (isIPv6(hostname)) {
-                urlStr = protocol+"://[" + hostname + "]:" + port + "/"; // IPv6 must be quoted in URLs
+                urlStr = protocol.getProtocol()+"://[" + hostname + "]:" + port + "/"; // IPv6 must be quoted in URLs
             }
             URI url = new URI(urlStr);
             return url.toASCIIString();
--- a/app/src/main/java/org/openobservatory/ooniprobe/common/ProxySettings.java
+++ b/app/src/main/java/org/openobservatory/ooniprobe/common/ProxySettings.java
@@ -33,8 +33,8 @@ public class ProxySettings {
             settings.protocol = ProxyProtocol.NONE;
         } else if (protocol.equals(ProxyProtocol.PSIPHON.getProtocol())) {
             settings.protocol = ProxyProtocol.PSIPHON;
-        } else if (protocol.equals(ProxyProtocol.SOCKS5.getProtocol())) {
-            settings.protocol = ProxyProtocol.SOCKS5;
+        } else if (protocol.equals(ProxyProtocol.SOCKS5.getProtocol()) || protocol.equals(ProxyProtocol.HTTP.getProtocol()) || protocol.equals(ProxyProtocol.HTTPS.getProtocol())) {
+            settings.protocol = ProxyProtocol.valueOf(protocol.toUpperCase());
         } else {
             // This is where we will extend the code to add support for
             // more proxies, e.g., HTTP proxies.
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/openobservatory/ooniprobe/common/ProxySettings.java
```diff
@@ -33,8 +33,8 @@ public static ProxySettings newProxySettings(PreferenceManager pm) throws Invali
             settings.protocol = ProxyProtocol.NONE;
         } else if (protocol.equals(ProxyProtocol.PSIPHON.getProtocol())) {
             settings.protocol = ProxyProtocol.PSIPHON;
-        } else if (protocol.equals(ProxyProtocol.SOCKS5.getProtocol())) {
-            settings.protocol = ProxyProtocol.SOCKS5;
+        } else if (protocol.equals(ProxyProtocol.SOCKS5.getProtocol()) || protocol.equals(ProxyProtocol.HTTP.getProtocol()) || protocol.equals(ProxyProtocol.HTTPS.getProtocol())) {
+            settings.protocol = ProxyProtocol.valueOf(protocol.toUpperCase());
```
bassosimone, line 37: Why are you using uppercase here?

app/src/main/java/org/openobservatory/ooniprobe/common/ProxySettings.java
```diff
@@ -72,21 +72,22 @@ private boolean isIPv6(String hostname) {
 
     /** getProxyString returns to you the proxy string you should pass to oonimkall. */
     public String getProxyString() throws URISyntaxException {
-        if (protocol == ProxyProtocol.NONE)
+        if (protocol == ProxyProtocol.NONE) {
             return "";
-        if (protocol == ProxyProtocol.PSIPHON)
+        } else if (protocol == ProxyProtocol.PSIPHON) {
```
bassosimone, line 77: You don't need `else` after return

app/src/main/java/org/openobservatory/ooniprobe/common/ProxySettings.java
```diff
@@ -72,21 +72,22 @@ private boolean isIPv6(String hostname) {
...
             return "psiphon://";
-        if (protocol == ProxyProtocol.SOCKS5) {
+        } else if (protocol == ProxyProtocol.SOCKS5||protocol == ProxyProtocol.HTTP||protocol == ProxyProtocol.HTTPS) {
```
bassosimone, line 79: You don't need `else` after return

app/src/main/java/org/openobservatory/ooniprobe/common/ProxySettings.java
```diff
@@ -72,21 +72,22 @@ private boolean isIPv6(String hostname) {
...
-            String urlStr = "socks5://" + hostname + ":" + port + "/";
+            String urlStr = protocol+"://" + hostname + ":" + port + "/";
             if (isIPv6(hostname)) {
-                urlStr = "socks5://[" + hostname + "]:" + port + "/"; // IPv6 must be quoted in URLs
+                urlStr = protocol+"://[" + hostname + "]:" + port + "/"; // IPv6 must be quoted in URLs
             }
             URI url = new URI(urlStr);
             return url.toASCIIString();
+        } else {
+            return "";
```
bassosimone, line 89: You don't need this else because the previous `if` branch is already return terminated

app/src/main/java/org/openobservatory/ooniprobe/common/ProxySettings.java
```diff
@@ -72,21 +72,22 @@ private boolean isIPv6(String hostname) {
...
             return "";
-        if (protocol == ProxyProtocol.PSIPHON)
+        } else if (protocol == ProxyProtocol.PSIPHON) {
             return "psiphon://";
-        if (protocol == ProxyProtocol.SOCKS5) {
+        } else if (protocol == ProxyProtocol.SOCKS5||protocol == ProxyProtocol.HTTP||protocol == ProxyProtocol.HTTPS) {
             // Alright, we now need to construct a new SOCKS5 URL. We are going to defer
             // doing that to the Java standard library (er, the Android stdlib).
-            String urlStr = "socks5://" + hostname + ":" + port + "/";
+            String urlStr = protocol+"://" + hostname + ":" + port + "/";
```
bassosimone, line 82: My concern here is that, while `HTTP` and `http` should be equivalent as URL schemes, I'd rather make sure we are using the lowercase version of the scheme, since it's more canonical and I wonder whether using the uppercase version may eventually lead to bugs downstream.

bassosimone (changes requested):

app/src/main/java/org/openobservatory/ooniprobe/common/ProxySettings.java
aanorbel: ProxyProtocol.valueOf will only accept one of the values in ProxyProtocol as in the enum definition(uppercase).

PR #627, opened 2023-10-24
Upgrade gradle to `8.1.2` and add support for kotlin and `kapt`

aanorbel:
## Proposed Changes

  - Upgrade `gradle` to `8.1.2`
  - Add support for `kotlin` and `kapt` to serve as base for writing kotlin code moving forward


bassosimone (approved):
🐳 but a test is failing and we should take a look

aanorbel:
Some tests fail randomly. its an issue I need to look into.

PR #628, opened 2023-10-25
Chore: Update from `ru.noties:markwon:2.0.1` to `io.noties.markwon:core:4.6.2`

aanorbel:
## Proposed Changes

  - Update from `ru.noties:markwon:2.0.1` to `io.noties.markwon:core:4.6.2`
  - Update required API usage.

Note: This update will allow us to use the `markwon` [plugin system](https://noties.io/Markwon/docs/v4/core/plugins.html) to implement additional functionality like the [Read more](https://github.com/noties/Markwon/blob/v4.6.2/app-sample/src/main/java/io/noties/markwon/app/samples/ReadMorePluginSample.java#L48C7-L208) which is added as part of the new designs.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -122,7 +122,7 @@ dependencies {
 
     implementation libs.xanscale.localhost.toolkit
     implementation libs.lottie
-    implementation libs.markwon
+    implementation libs.markwon.core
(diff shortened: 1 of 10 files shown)


bassosimone (approved):
🐳

PR #631, opened 2023-10-30
[New Designs] Update Dashboard view to match new design

aanorbel:
Fixes https://github.com/ooni/probe/issues/2588

## Proposed Changes

  -
  -
  -


aanorbel:
@agrabeli , We have been showing `test_name` on the  `test progress  fragment` but the new designs say we use the `test suite` name in the copy. I think we need a new copy for this section which may be different from the designs. What do you think.

agrabeli:
Thanks @aanorbel for your question (though I feel like I may be missing some context). Is the goal to display (for example) `Websites running` in the test progress section, as opposed to `Web Connectivity running`? If that is the case, what new copy would be needed? Can we use the existing test suite copy (`Websites`, `Instant Messaging`, etc), or would we need new strings for `Websites running`, `Instant Messaging running`, etc?

Overall, why is this change needed? While I can see that perhaps having "obscure" experiment names is not preferable from a UX perspective, at the same time it may be more interesting to see (for example) that "the WhatsApp test is running, the Signal test is running..." than "Instant Messaging running"...

aanorbel:
The new design calls for the test suite to be used instead of the test name. Wanted to know which side you're on.

agrabeli:
@aanorbel I'd personally prefer keeping the test names in the progress bar as I personally like seeing which experiment is running, and which experiment doesn't run (for example, due to network issues etc). I also find it more informative seeing that (for example) "WhatsApp is running" than "Instant Messaging is running" (though I realize that not everyone is familiar with the test names, and that some test names are not very descriptive). But if the consensus is to replace the test names with the test suites, we could do that.

app/build.gradle
```diff
@@ -103,6 +103,7 @@ dependencies {
 
 // AndroidX
     implementation libs.androidx.appcompat
+    //implementation libs.androidx.core
```
bassosimone, line 106: Please, don't keep commented out code around unless there's a reason.

app/src/main/java/org/openobservatory/ooniprobe/activity/MainActivity.java
```diff
@@ -85,7 +93,17 @@ protected void onCreate(@Nullable Bundle savedInstanceState) {
                         return false;
                 }
             });
+			// TODO(aanorbel): Fix change in state(theme change from notification) changing the selected item.
```
bassosimone, line 96: Is this a TODO for the current PR or for the future?

app/src/main/java/org/openobservatory/ooniprobe/adapters/DashboardAdapter.kt
```diff
@@ -48,15 +48,17 @@ class DashboardAdapter(
         val item = items[position]
         when (holder.itemViewType) {
             VIEW_TYPE_TITLE -> {
-            }
+				val separator = holder as CardGroupTitleViewHolder
+				separator.binding.root.text = item as String
```
bassosimone, line 52: beware that here it seems something is wrong with the indentation

app/src/main/java/org/openobservatory/ooniprobe/fragment/ProgressFragment.kt
```diff
@@ -0,0 +1,199 @@
...
+			service?.task?.let { task ->
+				task.currentSuite?.let {
+					biding.progress.max = service.task.getMax(preferenceManager)
+				}
+				task.currentTest?.let { currentTest ->
+					when {
+						/**
+						 * If the test has no icon, use the suite icon.
+						 * Currently not used since the most suites have a larger than desired icon.
+						 */
```
bassosimone, line 104: Thanks! This makes it clear why the code is commented out.

app/src/main/java/org/openobservatory/ooniprobe/fragment/ProgressFragment.kt
```diff
@@ -0,0 +1,199 @@
...
+		override fun onProgress(state: Int, eta: Double) {
+			updateUI(receiver.service)
+
+			biding.progress.apply {
+				isIndeterminate = false
+				progress = state
+			}
+		}
+
+		override fun onLog(value: String) {/* nothing */
```
bassosimone, line 179: Typically I find this pattern better:

```Kotlin
override fun onLog(value: String) {
  /* nothing */
}
```

bassosimone (changes requested):
Just minor changes

app/src/main/java/org/openobservatory/ooniprobe/activity/MainActivity.java
aanorbel: Future work

bassosimone (approved):
🐳

PR #637, opened 2023-11-26
Chore: Migrate dashboard and overview to descriptors

aanorbel:
## Proposed Changes

  - Add `OONIDescriptor` and methods to transform to `DynamicSuite` 
  - Update various components to use descriptors
  - Update tests


app/src/main/java/org/openobservatory/ooniprobe/model/database/Result.java
```diff
@@ -221,4 +202,23 @@ public void delete(Context c) {
...
+                    test_group_name,
+                    test_group_name,
+                    test_group_name,
+                    test_group_name,
+                    R.color.color_gray7_1,
+                    test_group_name,
+                    R.string.TestResults_NotAvailable,
+                    new ArrayList<>(),
+                    new ArrayList<>()
+            );
```
bassosimone, line 221: The indenting here is off. Also I do not fully understand what you're doing here. Are you returning the empty base nettest in this case? What is the implication of returning it? When could the exception occur?

app/src/main/java/org/openobservatory/ooniprobe/test/suite/AbstractSuiteExtensions.kt
```diff
@@ -1,56 +1,35 @@
...
+                    AbstractTest.getTestByName(test.name).apply {
+                        setOrigin(origin)
+                        inputs = urls
+                    }
+                )
+                return suite
+            }
+        }
+    }
+    return null
```
bassosimone, line 34: 🎻

app/src/main/java/org/openobservatory/ooniprobe/test/suite/README.md
```diff
@@ -0,0 +1,17 @@
...
+- [InstantMessagingSuite](https://github.com/ooni/probe-android/blob/7b8ec932fbccbb4f89e8101fddbe27d5a5591bf2/app/src/main/java/org/openobservatory/ooniprobe/test/suite/InstantMessagingSuite.java)
+- [CircumventionSuite](https://github.com/ooni/probe-android/blob/7b8ec932fbccbb4f89e8101fddbe27d5a5591bf2/app/src/main/java/org/openobservatory/ooniprobe/test/suite/CircumventionSuite.java)
+- [PerformanceSuite](https://github.com/ooni/probe-android/blob/7b8ec932fbccbb4f89e8101fddbe27d5a5591bf2/app/src/main/java/org/openobservatory/ooniprobe/test/suite/PerformanceSuite.java)
+- [ExperimentalSuite](https://github.com/ooni/probe-android/blob/7b8ec932fbccbb4f89e8101fddbe27d5a5591bf2/app/src/main/java/org/openobservatory/ooniprobe/test/suite/ExperimentalSuite.java)
+
+These test suites were removed and replaced with a single test suite
+called [`DynamicTestSuite`](./DynamicTestSuite.kt).
+
+The reason for this change is that we want to be able to dynamically build the `TestSuite` based on
+a `Descriptor` and also decouple the actual files on disk from the UI that are available in the app.
```
bassosimone, line 17: 🎻

bassosimone (commented):
It mostly LGTM. I would like some extra explanations in some places in the code before moving forward and merging. Thank you!

app/src/main/java/org/openobservatory/ooniprobe/activity/customwebsites/CustomWebsiteActivity.kt
```diff
@@ -102,12 +112,16 @@ class CustomWebsiteActivity : AbstractActivity(), ConfirmDialogFragment.OnClickL
...
+        val suite = descriptorManager.getTestByDescriptorName(OONITests.WEBSITES.label)
+        suite?.let {
+            suite.getTestList(preferenceManager)[0].inputs = urls
+            RunningActivity.runAsForegroundService(
+                this@CustomWebsiteActivity, suite.asArray(), { finish() }, preferenceManager
+            )
+            return true
+        }.run {
+            return false
+        }
```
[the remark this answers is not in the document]
aanorbel: `.run` block is for when the suite is null. The caller requires a return type depending on the success of the operation

app/src/main/java/org/openobservatory/ooniprobe/common/service/ServiceUtil.java
```diff
@@ -75,17 +75,18 @@ public static void startRunTestServiceUnattended(Application app) {
...
         AbstractSuite suite = d.generateAutoRunServiceSuite.generate();
         ArrayList<AbstractSuite> testSuites = new ArrayList<>();
         testSuites.add(suite);
-        testSuites.add(InstantMessagingSuite.initForAutoRun());
-        testSuites.add(CircumventionSuite.initForAutoRun());
-        testSuites.add(PerformanceSuite.initForAutoRun());
-        testSuites.add(ExperimentalSuite.initForAutoRun());
+        testSuites.addAll(Lists.transform(List.of(OONITests.INSTANT_MESSAGING, OONITests.CIRCUMVENTION, OONITests.PERFORMANCE, OONITests.EXPERIMENTAL), item -> {
+            DynamicTestSuite testSuite = item.toOONIDescriptor(app).getTest(app);
+            testSuite.setAutoRun(true);
```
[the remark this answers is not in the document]
aanorbel: Yes correct

app/src/main/java/org/openobservatory/ooniprobe/activity/customwebsites/CustomWebsiteActivity.kt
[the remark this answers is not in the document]
bassosimone: Thanks!

app/src/main/java/org/openobservatory/ooniprobe/model/database/Result.java
bassosimone: I would recommend to use a more idiomatic pattern. Return an optional type and make it null in this case, then modify the caller to check whether the returned value is empty. This feels better than returning the zero value object because it's more robust in that it reduces the coupling between parts of the codebase.

aanorbel: Let me do that instead. I didn't want to do so many null checks in java. 🤣

aanorbel: On second thought, i dont think we should change the current way it works

bassosimone: Oh, right, the consumer is still written in Java 🤦 

See, I was starting to think in Kotlin!

aanorbel: Right now, It works by using the enum as a lookup. Once descriptors are saved in the database, there is still a possibility that the test name will not be found. to avoid displaying errors and blank screens, it is better to have the test name in all areas where any other test information about that test is to be displayed.

bassosimone (approved):
🐳 🎻

PR #644, opened 2023-12-12
Release: 3.8.5.1, Update version code for `3.8.5.1` release

aanorbel:


app/build.gradle
```diff
@@ -13,8 +13,8 @@ android {
         applicationId 'org.openobservatory.ooniprobe'
         minSdk libs.versions.minSdk.get().toInteger()
         targetSdk libs.versions.targetSdk.get().toInteger()
-        versionName '3.8.5'
-        versionCode 107
+        versionName '3.8.5.1'
```
bassosimone, line 16: Can you make it 3.8.6? I am not 100% sure we should deviate from three version codes and semver.

bassosimone (changes requested):

app/build.gradle
aanorbel: I am not. this patch is already live for a quick fix so I don't know how that would go.

PR #647, opened 2023-12-14
Chore: Add support for sending build to firebase app distribution

aanorbel:
Fixes https://github.com/ooni/probe/issues/2643

## Proposed Changes

  - Adds Credentials to the repo
  - Add a new workflow file for Firebase App Distribution which publishes debug `apk`.


bassosimone (approved):
🐳

PR #648, opened 2023-12-18
Fix: Updated Broadcast receiver to do nothing if key is null

aanorbel:


app/src/main/java/org/openobservatory/ooniprobe/receiver/TestRunBroadRequestReceiver.java
```diff
@@ -49,6 +49,10 @@ public TestRunBroadRequestReceiver(PreferenceManager preferenceManager, EventLis
     public void onReceive(Context context, Intent intent) {
         String key = intent.getStringExtra("key");
         String value = intent.getStringExtra("value");
+        // If either key is null, do nothing
```
bassosimone, line 52: Here you say "either": which other key should we be checking below?

aanorbel: None.

aanorbel, line 52:
```suggestion
        // If key is null, do nothing
```

app/src/main/java/org/openobservatory/ooniprobe/common/service/RunTestService.java
```diff
@@ -209,6 +209,9 @@ public void onReceive(Context context, Intent intent) {
             ) {
                 return;
             }
+            if (key == null) {
```
aanorbel, line 212:
```suggestion
            // If key is null, do nothing
            if (key == null) {
```

bassosimone (approved):
🐳

PR #650, opened 2024-01-05
Feat: Add task to build experimental engine in android

aanorbel:
Fixes # <issue>

## Proposed Changes

  - Add a Gradle task to build `experimental-engine` from Android App.
  - Ensure the clean task deletes the existing engine archive.

```shell
./gradlew clean buildExperimentalArchive 
```

### DIFF
--- a/.github/workflows/build_experimental_engine.yml
+++ b/.github/workflows/build_experimental_engine.yml
@@ -26,7 +26,7 @@ jobs:
 
       - name: Get GOVERSION content
         id: goversion
-        run: echo "version=$(cat GOVERSION)" >> "$GITHUB_OUTPUT"
+        run: echo "version=$(curl https://raw.githubusercontent.com/ooni/probe-cli/master/GOVERSION && cat GOVERSION)" >> "$GITHUB_OUTPUT"
         shell: bash
 
       - uses: magnetikonline/action-golang-cache@v4
--- a/.github/workflows/build_experimental_engine.yml
+++ b/.github/workflows/build_experimental_engine.yml
@@ -32,7 +32,7 @@ jobs:
       - uses: magnetikonline/action-golang-cache@v4
         with:
           go-version: "${{ steps.goversion.outputs.version }}"
-          cache-key-suffix: "-coverage-${{ steps.goversion.outputs.version }}"
+          cache-key-suffix: "-build-experimental-engine-${{ steps.goversion.outputs.version }}"
 
       - name: Build Experimental Archive from `ooni/probe-cli`
         run: ./gradlew buildExperimentalArchive
--- a/.github/workflows/build_experimental_engine.yml
+++ b/.github/workflows/build_experimental_engine.yml
@@ -1,9 +1,15 @@
-name: build_experimental_engine
+name: Build Experimental Engine
 
-on: [push]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  schedule:
+    - cron: "0 2 * * */2"
 
 jobs:
-  build_android:
+  build_experimental_engine:
     runs-on: ubuntu-22.04
 
     steps:
@@ -30,3 +36,12 @@ jobs:
 
       - name: Build Experimental Archive from `ooni/probe-cli`
         run: ./gradlew buildExperimentalArchive
+
+      - name: Build Experimental App
+        run: ./gradlew assembleExperimentalFullDebug
+
+      - name: Archive Experimental APK
+        uses: actions/upload-artifact@v3
+        with:
+          name: app-experimental-full-debug
+          path: app/build/outputs/apk/experimentalFull/debug/app-experimental-full-debug.apk
--- a/.github/workflows/build_experimental_engine.yml
+++ b/.github/workflows/build_experimental_engine.yml
@@ -1,0 +1,32 @@
+name: build_experimental_engine
+
+on: [push]
+
+jobs:
+  build_android:
+    runs-on: ubuntu-22.04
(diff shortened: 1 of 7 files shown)


scripts/build_experimental_archive.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/sh
```
bassosimone, line 1: I suggest to include the following line right at the beginning:

```suggestion
#!/bin/sh
set -euxo pipefail
```

Here's what it does:

1) `-e` terminates the script at the first error

2) `-u` terminates the script if you use any undefined variable

3) `-x` prints each command that is executed (very useful for debugging)

4) `-o pipefail` causes a pipeline to fail if any command inside a pipeline fails

scripts/build_experimental_archive.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Define build and output directories
+gitTarget=${1:-"master"}
+buildDir=${2:-"./build"}
+
+# Create build directory and navigate into it
+mkdir -p "${buildDir}" && cd "${buildDir}"
```
bassosimone, line 8: You don't need the `&&` with `set -euxo pipefail`.

Also, I'd `rm -r ${buildDir}` before entering into it, otherwise `git clone` would fail the second time you run it. If you instead want to keep the `${buildDir}` and the cloned repository around, you need logic to avoid cloning the repository if it's already there. In such a case, you would need to pull the sources again.

Evidence:

```text
% git clone -v https://github.com/ooni/probe-cli.git
Cloning into 'probe-cli'...
POST git-upload-pack (175 bytes)
POST git-upload-pack (gzip 8652 to 4314 bytes)
remote: Enumerating objects: 29352, done.
remote: Counting objects: 100% (332/332), done.
remote: Compressing objects: 100% (220/220), done.
remote: Total 29352 (delta 141), reused 247 (delta 103), pack-reused 29020
Receiving objects: 100% (29352/29352), 29.08 MiB | 9.65 MiB/s, done.
Resolving deltas: 100% (20037/20037), done.

% git clone -v https://github.com/ooni/probe-cli.git
fatal: destination path 'probe-cli' already exists and is not an empty directory.
```

scripts/build_experimental_archive.sh
```diff
@@ -0,0 +1,29 @@
...
+
+# Clone the repository
+git clone -v https://github.com/ooni/probe-cli.git
+
+# Navigate into the repository
+cd probe-cli
+
+git checkout master
```
bassosimone, line 16: If you only want to clone each time, this line is redundant. Otherwise, I think it's fine to have it, but you would still need to add logic above to handle the case where the repository already exists.

scripts/build_experimental_archive.sh
```diff
@@ -0,0 +1,29 @@
...
+
+git pull origin
+# Checkout the target branch
+git checkout -c "${gitTarget}"
+
+# Build the probe-cli
+make android
```
bassosimone, line 23: This command has huge requirements in terms of building, but I reckon many of them are probably already satisfied if one is using Android Studio every day. Trying to spell them out, I think they are:

* the correct version of Java being installed (is that Java 17 now?)
* `ANDROID_HOME` being defined
* the `ANDROID_HOME` must contain the correct version of the NDK and the SDK
* the correct version of Go being installed
* the C/C++ compiler being installed

I think most of these preconditions are checked by `make android`, but I also think we should create a GitHub action installing all the required prerequisites, which would serve two purposes:

1. giving us confidence that this method of building works
2. documenting what are the actual pre-requisites

This action does not need to run at every pull request, rather I suppose it's fine to run it only for commits that have already been merged into the `master` branch.

bassosimone (changes requested):
Thank you, this is a good starting point, but we need some extra work to (a) define the usage pattern (clone each time versus keeping the cloned repo) and (b) add some integration tests to make sure all of this is WAI.

.github/workflows/build_experimental_engine.yml
```diff
@@ -0,0 +1,47 @@
...
+
+      - name: Get GOVERSION content
+        id: goversion
+        run: echo "version=$(curl https://raw.githubusercontent.com/ooni/probe-cli/master/GOVERSION && cat GOVERSION)" >> "$GITHUB_OUTPUT"
+        shell: bash
+
+      - uses: magnetikonline/action-golang-cache@v4
+        with:
+          go-version: "${{ steps.goversion.outputs.version }}"
+          cache-key-suffix: "-coverage-${{ steps.goversion.outputs.version }}"
```
bassosimone, line 35:
```suggestion
          cache-key-suffix: "-build-experimental-engine-${{ steps.goversion.outputs.version }}"
```

.github/workflows/build_experimental_engine.yml
```diff
@@ -0,0 +1,47 @@
...
+          java-version: '17'
+          distribution: 'temurin'
+
+      - name: Setup Android SDK
+        uses: android-actions/setup-android@v3
+
+      - name: Get GOVERSION content
+        id: goversion
+        run: echo "version=$(curl https://raw.githubusercontent.com/ooni/probe-cli/master/GOVERSION && cat GOVERSION)" >> "$GITHUB_OUTPUT"
+        shell: bash
```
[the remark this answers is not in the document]
bassosimone: We don't need this anymore as soon as https://github.com/ooni/probe/issues/2664 is closed. As long as Go >= 1.19 is installed, `make android` would install and use the right version of Go for building.

PR #654, opened 2024-01-17
Update logic to determine preference values

aanorbel:
Fixes OONI cards not enabled by default.

## Proposed Changes

  - Set default value for ooni tests to enabled.


bassosimone (approved):
🐳

PR #658, opened 2024-01-26
chore: Update android gradle plugin, preference, lifecycle process, gson, material and firebase dependencies

aanorbel:
## Proposed Changes

  - Update android gradle plugin, preference, lifecycle process, gson, material and firebase dependencies.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -139,7 +139,7 @@ dependencies {
     fullImplementation libs.google.play.core
 
 // Dependency Injection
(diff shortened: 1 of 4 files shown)


bassosimone (approved):
🐳

bassosimone:
The test suite is still throwing out of memory but maybe this branch does not include recent changes to make sure the tests are working as intended by, among other things, bumping the memory used by the JVM?

aanorbel:
Sure I will do that.

PR #659, opened 2024-01-30
chore: Add dependency bot and update `sentry` and `countly`

aanorbel:
## Proposed Changes

  - Add `dependency bot` configuration.
  - Update  `sentry` and `countly` SDK which the IDE cannot determine if updates are present.


bassosimone (approved):
🐳

PR #673, opened 2024-02-20
fix: Improve how in progress tests are displayed

aanorbel:
Fixes  https://github.com/ooni/probe/issues/2658

## Proposed Changes

  -
  -
  -

### DIFF
--- a/app/src/main/java/org/openobservatory/ooniprobe/model/database/ResultExtensions.kt
+++ b/app/src/main/java/org/openobservatory/ooniprobe/model/database/ResultExtensions.kt
@@ -1,0 +1,44 @@
+package org.openobservatory.ooniprobe.model.database
+
+import com.raizlabs.android.dbflow.config.FlowManager
+import com.raizlabs.android.dbflow.structure.database.FlowCursor
+import org.openobservatory.ooniprobe.common.AppDatabase
+
+
+class ResultExtensions {
+
+    companion object {
+
+        @JvmStatic
+        fun Result.getStatus(): Int {
+            val queryById = """
+        SELECT
+            CASE 
+                WHEN COUNT(m.result_id) = 0 THEN -1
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/openobservatory/ooniprobe/test/TestAsyncTask.java
```diff
@@ -151,6 +151,10 @@ private void runTest(AbstractTest... tests) {
                     currentTest.run(app, app.getPreferenceManager(),app.getLogger(), app.getGson(), result, i, this);
                 }
             }
+            if (result.countTotalMeasurements() == tests.length) {
```
hellais, line 154: Where is this function defined?

Maybe this function should be called "countTotalCompletedMeasurements" or something of that sort.

Also, what exactly is counted here? Is it really a measurements or is it a nettest in a group?

For example if you run `web_connectivity` against 10 URLs and then `whatsaapp`, would this count be 11 or would it be 2?

app/src/main/java/org/openobservatory/ooniprobe/model/database/ResultExtensions.kt
```diff
@@ -0,0 +1,44 @@
...
+    """.trimIndent()
+            val cursor: FlowCursor = FlowManager.getDatabase(AppDatabase::class.java)
+                .writableDatabase
+                .rawQuery(queryById, arrayOf(id.toString()))
+
+            var inProgressCount = 0
+            if (cursor.moveToFirst()) {
+                cursor.getColumnIndex("in_progress_count").let {
+                    if (it == -1) {
+                        return@let
```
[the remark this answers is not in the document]
aanorbel: Exit the let call and not the function

app/src/main/java/org/openobservatory/ooniprobe/test/TestAsyncTask.java
aanorbel: 11

PR #685, opened 2024-02-29
feat: Switched runner to ubuntu and update actions

aanorbel:
## Proposed Changes

  - Switched runner to Ubuntu
  - Update actions to the latest version


aanorbel (commented):
🐳

PR #713, opened 2024-04-02
feat (OONI Run v2): Change backend for fetching descriptors

aanorbel:
## Proposed Changes

  - Change the backend for fetching descriptors and refactor code appropriately.


engine/src/main/java/org/openobservatory/engine/OONISessionConfig.java
```diff
@@ -39,7 +39,7 @@ public final class OONISessionConfig {
      * usage of an alternative probe services. This setting
      * should only be used to implement integration tests.
      */
-    String probeServicesURL;
+    public String probeServicesURL;
```
bassosimone, line 42: Unless I'm missing something, the probe services URL is always passed as an argument and therefore we can back out this specific change.

```suggestion
    String probeServicesURL;
```

PR #767, opened 2024-06-28
build: enable API desugaring

Lanius-collaris:
commons-io 2.15.1 requires `public java.nio.file.Path toPath()`, but this api is available since Android 8.0, so I enabled API desugaring.

Fixes ooni/probe#2750


aanorbel:
Clossed by https://github.com/ooni/probe-android/pull/789

PR #782, opened 2024-07-16
Fix: Instrumented tests on CI

sdsantos:
Fixes https://github.com/ooni/probe/issues/2779

## Proposed Changes
  - Enable hardware accelerated emulators (KVM) as suggested on https://github.com/ReactiveCircus/android-emulator-runner
  - Switch the emulator profile to example given on https://github.com/ReactiveCircus/android-emulator-runner
  - Remove the `disable-animations` because the default value is already `true`


aanorbel (commented):
🐳

PR #784, opened 2024-07-16
Fix: OONIAPIClientTest using failing API endpoint and incorrect testing logic

sdsantos:
Fixes https://github.com/ooni/probe/issues/2778

## Proposed Changes

  - Switch testing endpoint from `ams-pg.ooni.org` to `https://backend-hel.ooni.org`
  - Refactor `OONIAPIClientTest` to Kotlin
  - Execute API calls synchronously to fix incorrect async logic
  - Remove database operations that were irrelevant to the `testSelectMeasurementsWithJson` test
  - Remove the test `getMeasurementJsonError` because it was only testing external libraries logic (retrofit and gson) and not the `OONIAPIClient`


aanorbel (approved):
🐳

PR #785, opened 2024-07-17
Fix flaky ResultTest and build config warnings

sdsantos:
I haven't created a ticket for these changes.

## Proposed Changes

  - Fix `ResultTest.getLastResult` by ensuring that the first result really has a start_time before the second result
  - Fix `BuildConfig` warning by changing to the new configuration
  - Remove the `mockito-inline` dependency, since it's no longer needed for Mockito 5

### Context

- This PR started as an attempt to fix https://github.com/ooni/probe/issues/2777 but I wasn't able to do it yet

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -117,6 +117,7 @@ android {
     }
     buildFeatures {
         viewBinding = true
+        buildConfig = true
     }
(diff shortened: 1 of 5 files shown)


aanorbel (approved):
🐳

PR #852, opened 2024-11-21
fix(engine): expose session close

DecFox:
Fixes https://github.com/ooni/probe/issues/2810

## Proposed Changes

  - Exposes the `close` function for the engine session 
  - Invokes the `close` function while downloading urls for webconnectivity to ensure the running psiphon tunnel (and engine session) is closed before the experiment run
  - Changes in engine api surface: https://github.com/ooni/probe-cli/pull/1657


aanorbel:
@DecFox , we need to update the engine version to ensure the build works before we can merge this.

